PR #3, opened 2019-11-21
Black

vkhromov:


drmorr0:
I... don't really want black on Clusterman.  I haven't had great experiences with it thus far.  :(

PR #4, opened 2019-12-11
Reload Kube Configs

chlgit:
I read the kubernetes client library code again. Now, I believe watch pki keys is not enough because it caches the token here https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L646. I think we should reload the kube configs. I've tested it at stagef successfully.


drmorr0:
This seems fine but you'll also need to add the watchers for the metrics collector batch.  I suggest refactoring this code into `batch/util.py`, where we have some other similar code set up.

drmorr0 (approved):

PR #5, opened 2019-12-11
Bump yelp-batch to 10.0.2

vkhromov:


vkhromov:
`make test`/`make itest` internal runs were successful as well.

drmorr0 (approved):

PR #6, opened 2019-12-27
add scheduler attribute for instances

chlgit:
I have tested this at stagef and fairly sure it is backward compatible.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -56,6 +56,7 @@ class Host(NamedTuple):
     ip: str
     sender: str
     receipt_handle: str
+    scheduler: str = 'mesos'
 
 
 class DrainingClient():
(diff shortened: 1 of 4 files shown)


clusterman/draining/queue.py
```diff
@@ -312,6 +321,10 @@ def host_from_instance_id(
         return None
     try:
         sfr_ids = [tag['Value'] for tag in instance_data[0]['Tags'] if tag['Key'] == 'aws:ec2spot:fleet-request-id']
+        scheduler = 'mesos'
+        for tag in instance_data[0]['Tags']:
+            if tag['Key'] == 'KubernetesCluster':
+                scheduler = 'kubernetes'
```
drmorr0, line 327: nit: you could break after this line, right?

drmorr0 (approved):
fix'n'ship

PR #7, opened 2019-12-28
Prevent connection errrors from crashing the autoscaler

kawaiwanyelp:
### Description
We've had some flappy alerts whenever Clusterman crashes when it fails to contact a Mesos master, causing a `clusterman.exceptions.PoolManagerError`, or talk to AWS, causing a `botocore.exceptions.EndpointConnectionError, due to a DNS error. These alerts don't really mean much since Clusterman will just try to autoscale again after crashing, and the alerts go away once DNS rises like a mangled phoenix. I've made it so that it doesn't crash on these errors.

### Testing done
make test

### DIFF
--- a/clusterman/batch/autoscaler.py
+++ b/clusterman/batch/autoscaler.py
@@ -14,6 +14,7 @@
 import time
 
 import colorlog
+from botocore.exceptions import EndpointConnectionError
 from yelp_batch.batch import batch_command_line_arguments
 from yelp_batch.batch import batch_configure
 from yelp_batch.batch_daemon import BatchDaemon
@@ -31,6 +32,7 @@ from clusterman.batch.util import suppress_request_limit_exceeded
 from clusterman.config import setup_config
 from clusterman.exceptions import AutoscalerError
 from clusterman.exceptions import ClustermanSignalError
+from clusterman.exceptions import PoolConnectionError
 from clusterman.util import get_autoscaler_scribe_stream
 from clusterman.util import sensu_checkin
 from clusterman.util import setup_logging
@@ -132,7 +134,10 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
 
     def run(self):
         while self.running:
-            self._autoscale()
+            try:
+                self._autoscale()
+            except (PoolConnectionError, EndpointConnectionError) as e:
+                logger.exception(f'Encountered a connection error: {e}')
 
     def _do_sensu_checkins(self, signal_failed, service_failed, msg):
         check_every = ('{minutes}m'.format(minutes=int(self.autoscaler.run_frequency // 60))
--- a/clusterman/mesos/util.py
+++ b/clusterman/mesos/util.py
@@ -20,7 +20,7 @@ import colorlog
 import requests
 from mypy_extensions import TypedDict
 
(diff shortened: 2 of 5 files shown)


mzq592 (approved):
lgtm

clusterman/batch/autoscaler.py
```diff
@@ -132,7 +134,10 @@ def _autoscale(self):
 
     def run(self):
         while self.running:
-            self._autoscale()
+            try:
+                self._autoscale()
+            except (PoolConnectionError, EndpointConnectionError) as e:
+                logger.error(f'Encountered a connection error: {e}')
```
qui: `logger.exception` so we get the traceback

clusterman/mesos/util.py
```diff
@@ -103,7 +103,7 @@ def mesos_post(url: str, endpoint: str) -> requests.Response:
                 f'Response Text: {response.text}\n'
             )
         logger.critical(log_message)
-        raise PoolManagerError(f'Mesos master unreachable: check the logs for details') from e
+        raise PoolConnectionError(f'Mesos master unreachable: check the logs for details') from e
```
qui, line 106: Would it make sense to raise PoolConnectionError for a more limited type of exception, like `requests.exceptions.ConnectionError` ?

tests/batch/autoscaler_test.py
```diff
@@ -15,14 +15,16 @@
...
 
 try:
     from clusterman.batch.autoscaler import AutoscalerBatch
 except ImportError:
     pytest.mark.skip('Could not import the batch; are you in a Yelp-y environment?')
+from clusterman.exceptions import PoolConnectionError
 
 
-def batch(extra_args=None, mock_sensu=True):
+def batch(extra_args=[], mock_sensu=True):
```
qui: default arguments should not be mutable. You should keep `extra_args=None` and do this in the body instead:
`extra_args = extra_args or []`

extra reading if you haven't heard about this yet: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

tests/batch/autoscaler_test.py
```diff
@@ -75,3 +77,16 @@ def test_run(mock_sensu, mock_running, mock_time, mock_sleep, dry_run):
...
+@pytest.mark.parametrize('exc', [PoolConnectionError, EndpointConnectionError(endpoint_url='')])
+def test_run_connection_error(mock_running, exc):
+    batch_obj = batch()
+    batch_obj._autoscale = mock.Mock(side_effect=exc)
+    mock_running.side_effect = [True, True, False]
+
+    batch_obj.run()
+
+    # exceptions raised did prevent subsequent calls to autoscale
+    assert batch_obj._autoscale.call_count == 1
```
qui: I guess this test doesn't run in Travis (because it's not a yelpy env) but this should be 2, right?

clusterman/mesos/util.py
kawaiwanyelp: I don't think it does. There isn't a single unifying exception that is raised when we fail to reach Mesos, so the point of PoolConnectionError here is so that we can catch errors from mesos_post in _autoscale.

drmorr0 (approved):
please make sure you check the test @qui highlighted, otherwise shipit

qui (approved):

PR #12, opened 2020-01-21
taint node and evict pods for draining clusterman nodes

chlgit:
I've tested this at stagef, and it works fine.

If the node has been in draining queue, it would do 2 things currently:
  1) taint the node as no-schedule
  2) evict all pods running on this node
After that, it would be in termination queue for terminating.

Also, I don't expect to ship this until I go back from PTO. So take your time to review this.

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -100,8 +100,8 @@ class PoolManager:
             try:
                 group.mark_stale(dry_run)
             except NotImplementedError as e:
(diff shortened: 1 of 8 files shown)


enochtangg:
Since the `kube_operator_client` in this PR is useful for the instant deletion the kube node after EC2 termination, `u/enocht/CLUSTERMAN-468_delete_kube_node_on_ec2_terminate` has been rebased off this branch and is therefore dependent on this PR.

chlgit:
Close this in favor of @enochtangg's PR.

PR #13, opened 2020-01-23
reload API client to (hopefully) fix PKI issues

drmorr0:
I'm pretty sure it's insufficient to just reload the config, I think we have to reload the config and then create a new API client.  Hopefully this should resolve the pages we get about Clusterman credentials.

Also I fixed the clusterman/pip20 build problem.


mattmb (approved):
👍 lgtm

PR #15, opened 2020-01-30
pass through args from bootstrap to autoscaler

drmorr0:
This should make it easier to debug stuff and/or run the autoscaler manually with custom configs.


gvolco (approved):
lgtm

PR #16, opened 2020-01-30
don't terminate nodes with the safe-to-evict annotation

drmorr0:
After this change, you should be able to stick the `clusterman.com/safe_to_evict=false` annotation on a pod, and it will prevent the node that it's running on from getting terminated.

### DIFF
--- a/tests/kubernetes/kubernetes_cluster_connector_test.py
+++ b/tests/kubernetes/kubernetes_cluster_connector_test.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from copy import deepcopy
+
 import mock
 import pytest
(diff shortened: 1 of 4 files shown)


tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -50,53 +102,9 @@ def mock_cluster_connector():
...
-                    spec=V1PodSpec(containers=[
-                           V1Container(
-                                name='container1',
-                                resources=V1ResourceRequirements(requests={'cpu': '1.5'})
-                            )
-                        ]
-                    )
-                ),
-            ]
+            '10.10.10.2': [pod1, pod2],
```
stug, line 108: I might suggest having a case where there is a pod that is safe to evict on a node to show that that node is killable since that seems like a major case.

stug (approved):
lg2m

gvolco (approved):
lgtm!

PR #17, opened 2020-02-04
fix the "no annotations" bug

drmorr0:
This was causing crashes for pods that didn't have any annotations.

Confirmed that the change to the test fixes the issue.

Also made the cluster metrics batch produce a traceback, resolving CLUSTERMAN-457

### DIFF
--- a/clusterman/batch/cluster_metrics_collector.py
+++ b/clusterman/batch/cluster_metrics_collector.py
@@ -125,30 +125,35 @@ class ClusterMetricsCollector(BatchDaemon, BatchLoggingMixin, BatchRunningSentin
     def run(self) -> None:
         self.load_pool_managers()  # Load the pools on the first run; do it here so we get logging
 
-        while self.running:
-            time.sleep(splay_event_time(
-                self.run_interval,
-                self.get_name() + self.options.cluster,
-            ))
+        try:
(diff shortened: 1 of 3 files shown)


jangsutsr (approved):
lgtm

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -125,30 +125,35 @@ def load_pool_managers(self) -> None:
...
+                        check_every='1m',
+                        source=self.options.cluster,
+                        ttl='10m',
+                        noop=self.options.disable_sensu,
+                    )
+                    sensu_checkin(**sensu_args)
+
+        except Exception:
+            # yelp_batch doesn't show the whole traceback when something fails
+            self.logger.exception('cluster metrics collector failed')
```
stug, line 156: I'm assuming we'll get at least some traceback here?  Or will we just get this message?

stug (approved):

clusterman/batch/cluster_metrics_collector.py
drmorr0: self.logger.exception provides the traceback: https://docs.python.org/3/library/logging.html#logging.Logger.exception

PR #21, opened 2020-02-15
Add k8s draining and node deletion on terminate queue

enochtangg:
Description:
The following code is for CLUSTERMAN-468 and CLUSTERMAN-456. Currently, we only have any logic to do draining for mesos instances so this PR includes unscheduling and evicting k8s pods as part of a step in the draining queue. In addition, we also delete a k8s node immediately when the EC2 instance is terminated during the last step of the termination process queue. 

Testing Done:
* `make test`
* Unit tests
* End to end testing: Created clusterman autoscaler and drainer on kubestage and tested draining nodes with `clusterman manage`.

### DIFF
--- a/clusterman/draining/kubernetes.py
+++ b/clusterman/draining/kubernetes.py
@@ -1,0 +1,27 @@
+# Copyright 2019 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import colorlog
+
+logger = colorlog.getLogger(__name__)
+
+
+def kube_drain(operator_client, host):
+    operator_client.reload_state()
+    operator_client.set_node_unschedulable(host.ip)
+    operator_client.evict_pods_on_node(host.ip)
+
+
+def kube_delete_node(operator_client, host):
+    operator_client.reload_state()
+    operator_client.delete_node(host.ip)
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -37,11 +37,14 @@ from clusterman.aws.util import RESOURCE_GROUPS_REV
 from clusterman.config import load_cluster_pool_config
 from clusterman.config import POOL_NAMESPACE
 from clusterman.config import setup_config
+from clusterman.draining.kubernetes import kube_delete_node
+from clusterman.draining.kubernetes import kube_drain
 from clusterman.draining.mesos import down
 from clusterman.draining.mesos import drain
 from clusterman.draining.mesos import operator_api
 from clusterman.draining.mesos import up
 from clusterman.interfaces.resource_group import InstanceMetadata
+from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
 from clusterman.util import get_pool_name_list
 
 
@@ -226,13 +229,14 @@ class DrainingClient():
 
     def process_termination_queue(
         self,
-        mesos_operator_client: Callable[..., Callable[[str], Callable[..., None]]],
+        mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
+        kube_operator_client: Optional[KubernetesClusterConnector],
     ) -> None:
         host_to_terminate = self.get_host_to_terminate()
         if host_to_terminate:
             # as for draining if it has a hostname we should down + up around the termination
-            if host_to_terminate.hostname:
-                logger.info(f'Hosts to down+terminate+up: {host_to_terminate}')
+            if host_to_terminate.scheduler == 'mesos':
+                logger.info(f'Mesos hosts to down+terminate+up: {host_to_terminate}')
                 hostname_ip = f'{host_to_terminate.hostname}|{host_to_terminate.ip}'
                 try:
                     down(mesos_operator_client, [hostname_ip])
@@ -243,19 +247,27 @@ class DrainingClient():
                     up(mesos_operator_client, [hostname_ip])
                 except Exception as e:
                     logger.error(f'Failed to up {hostname_ip} continuing to terminate anyway: {e}')
+            elif host_to_terminate.scheduler == 'kubernetes':
+                logger.info(f'Kubernetes hosts to delete k8s node and terminate: {host_to_terminate}')
+                try:
+                    logger.info(f'Deleting kubernetes node')
+                    kube_delete_node(kube_operator_client, host_to_terminate)
+                except Exception as e:
+                    logger.warning(f'Failed to delete {host_to_terminate.ip} node continuing to terminate anyway: {e}')
+                terminate_host(host_to_terminate)
             else:
-                logger.info(f'Host to terminate: {host_to_terminate}')
+                logger.info(f'Host to terminate immediately: {host_to_terminate}')
                 terminate_host(host_to_terminate)
             self.delete_terminate_messages([host_to_terminate])
 
     def process_drain_queue(
         self,
-        mesos_operator_client: Callable[..., Callable[[str], Callable[..., None]]],
+        mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
+        kube_operator_client: Optional[KubernetesClusterConnector],
     ) -> None:
         host_to_process = self.get_host_to_drain()
         if host_to_process and host_to_process.instance_id not in self.draining_host_ttl_cache:
             self.draining_host_ttl_cache[host_to_process.instance_id] = arrow.now().shift(seconds=DRAIN_CACHE_SECONDS)
-
             if host_to_process.scheduler == 'mesos':
                 logger.info(f'Mesos host to drain and submit for termination: {host_to_process}')
                 try:
@@ -271,6 +283,7 @@ class DrainingClient():
                     self.submit_host_for_termination(host_to_process)
             elif host_to_process.scheduler == 'kubernetes':
                 logger.info(f'Kubernetes host to drain and submit for termination: {host_to_process}')
+                kube_drain(kube_operator_client, host_to_process)
                 self.submit_host_for_termination(host_to_process, delay=0)
             else:
                 logger.info(f'Host to submit for termination immediately: {host_to_process}')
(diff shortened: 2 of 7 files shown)


tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -132,13 +131,12 @@ def test_total_cpus(mock_cluster_connector):
...
-    assert len(mock_cluster_connector._get_pending_pods()) == 1
-
-
-def test_get_unschedulable_pods(mock_cluster_connector):
-    assert len(mock_cluster_connector.get_unschedulable_pods()) == 1
-
-
-def test_pending_cpus(mock_cluster_connector):
-    assert mock_cluster_connector.get_resource_pending('cpus') == 1.5
+def test_delete_node(mock_cluster_connector):
```
chlgit, line 157: nit: you might want to do the similar tests for evict_pods_on_node and set_node_unschedulable

clusterman/draining/queue.py
```diff
@@ -357,18 +372,24 @@ def host_from_instance_id(
 
 def process_queues(cluster_name: str) -> None:
     draining_client = DrainingClient(cluster_name)
-    mesos_master_url = staticconf.read_string(f'clusters.{cluster_name}.mesos_master_fqdn')
-    mesos_secret_path = staticconf.read_string(f'mesos.mesos_agent_secret_path', default=None)
-    operator_client = operator_api(mesos_master_url, mesos_secret_path)
+    cluster_manager_name = staticconf.read_string(f'clusters.{cluster_name}.cluster_manager')
+    mesos_operator_client = None
+    kube_operator_client = KubernetesClusterConnector(cluster_name, None)
```
chlgit: Will this cause problems for mesos-specific drainer? I think we can do something like:
try:
   setup operator client
except Errors:
   error handling

for both of kubernetes and mesos coperator clients.

clusterman/draining/queue.py
```diff
@@ -243,14 +247,24 @@ def process_termination_queue(
                     up(mesos_operator_client, [hostname_ip])
                 except Exception as e:
                     logger.error(f'Failed to up {hostname_ip} continuing to terminate anyway: {e}')
+            elif host_to_terminate.scheduler == 'kubernetes':
+                logger.info(f'Host to kubernetes terminate: {host_to_terminate}')
+                terminate_host(host_to_terminate)
+                if host_to_terminate.scheduler == 'kubernetes':
+                    try:
+                        logger.info(f'Deleting kubernetes node')
+                        kube_delete_node(kube_operator_client, host_to_terminate)
```
chlgit: Could we double check if we need to kube_delete_node before terminate_host? I worry the race condition here would cause users' error.

clusterman/draining/kubernetes.py
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+import colorlog
+
+logger = colorlog.getLogger(__name__)
+
+
+def kube_drain(operator_client, host):
+    operator_client.reload_state()
+    operator_client.set_node_unschedulable(host.ip)
+    operator_client.evict_pods_on_node(host.ip)
```
chlgit, line 22: Have you figured out how to solve the permission issue? Or do you want to log the error messages and figure it out later?

clusterman/autoscaler/pool_manager.py
```diff
@@ -178,6 +178,7 @@ def prune_excess_fulfilled_capacity(
                 assert self.draining_client  # make mypy happy
                 for group_id, node_metadatas in marked_nodes_by_group.items():
                     for node_metadata in node_metadatas:
+                        logger.info(f'Submitted {node_metadata} for draining')
```
chlgit: Do we need log entire metadata?

clusterman/autoscaler/pool_manager.py
```diff
@@ -437,8 +438,6 @@ def _get_prioritized_killable_nodes(self) -> List[ClusterNodeMetadata]:
     def _is_node_killable(self, node_metadata: ClusterNodeMetadata) -> bool:
         if node_metadata.agent.state == AgentState.UNKNOWN:
             return False
-        elif not node_metadata.agent.is_safe_to_kill:
```
drmorr0: Why did this get deleted?

clusterman/draining/queue.py
```diff
@@ -226,13 +229,14 @@ def delete_warning_messages(self, hosts: Sequence[Host]) -> None:
...
-        mesos_operator_client: Callable[..., Callable[[str], Callable[..., None]]],
+        mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
+        kube_operator_client: KubernetesClusterConnector,
     ) -> None:
         host_to_terminate = self.get_host_to_terminate()
         if host_to_terminate:
             # as for draining if it has a hostname we should down + up around the termination
-            if host_to_terminate.hostname:
-                logger.info(f'Hosts to down+terminate+up: {host_to_terminate}')
+            if not host_to_terminate.scheduler or host_to_terminate.scheduler == 'mesos':
```
drmorr0: maybe `if host_to_terminate.scheduler in {None, 'mesos'}:`?  Or maybe not, it's shorter but slightly less clear.

clusterman/draining/queue.py
```diff
@@ -243,14 +247,24 @@ def process_termination_queue(
                     up(mesos_operator_client, [hostname_ip])
                 except Exception as e:
                     logger.error(f'Failed to up {hostname_ip} continuing to terminate anyway: {e}')
+            elif host_to_terminate.scheduler == 'kubernetes':
+                logger.info(f'Host to kubernetes terminate: {host_to_terminate}')
+                terminate_host(host_to_terminate)
+                if host_to_terminate.scheduler == 'kubernetes':
```
drmorr0: Why are you checking the scheduler again here?

clusterman/draining/queue.py
```diff
@@ -226,13 +229,14 @@ def delete_warning_messages(self, hosts: Sequence[Host]) -> None:
...
+        mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
+        kube_operator_client: KubernetesClusterConnector,
     ) -> None:
         host_to_terminate = self.get_host_to_terminate()
         if host_to_terminate:
             # as for draining if it has a hostname we should down + up around the termination
-            if host_to_terminate.hostname:
-                logger.info(f'Hosts to down+terminate+up: {host_to_terminate}')
+            if not host_to_terminate.scheduler or host_to_terminate.scheduler == 'mesos':
+                logger.info(f'Hosts to mesos down+terminate+up: {host_to_terminate}')
```
drmorr0: Can you change this to say "Mesos hosts to down+terminate+up"?

clusterman/draining/queue.py
```diff
@@ -243,14 +247,24 @@ def process_termination_queue(
                     up(mesos_operator_client, [hostname_ip])
                 except Exception as e:
                     logger.error(f'Failed to up {hostname_ip} continuing to terminate anyway: {e}')
+            elif host_to_terminate.scheduler == 'kubernetes':
+                logger.info(f'Host to kubernetes terminate: {host_to_terminate}')
```
drmorr0: "kubernetes hosts to drain and terminate"

clusterman/draining/queue.py
```diff
@@ -243,14 +247,24 @@ def process_termination_queue(
...
+                terminate_host(host_to_terminate)
+                if host_to_terminate.scheduler == 'kubernetes':
+                    try:
+                        logger.info(f'Deleting kubernetes node')
+                        kube_delete_node(kube_operator_client, host_to_terminate)
+                    except Exception as e:
+                        logger.warning(f'Failed to delete {host_to_terminate.ip} node. Continuing anyway: {e}')
             else:
-                logger.info(f'Host to terminate: {host_to_terminate}')
+                logger.info(f'Host to regular terminate: {host_to_terminate}')
```
drmorr0: "Hosts to terminate immediately"

clusterman/autoscaler/pool_manager.py
enochtangg: This was a change made by @chlgit when implementing kube drain if you can chime in here.

clusterman/draining/queue.py
enochtangg: This ticket https://jira.yelpcorp.com/browse/CLUSTERMAN-468 specified that kube nodes are still active in scheduler even after `terminate_host`. My interpretation was that we needed to delete these nodes after the EC2 instance is terminated. Can @drmorr0 confirm this?

clusterman/draining/kubernetes.py
enochtangg: Not yet, but I've logged the warning errors within those functions `set_node_unschedulable` and `evict_pods_on_node`

clusterman/autoscaler/pool_manager.py
chlgit: @enochtangg I don't remember this and I don't see it from my [PR](https://github.com/Yelp/clusterman/pull/12/files). Could you double check this?

enochtangg: Weird, the changes I pulled had all references of `safe_to_kill` removed. I'll add it back it in, thanks!

clusterman/draining/queue.py
enochtangg: Update: you're correct, when attempting to delete the node off the k8s scheduler, the system threw errors because some nodes already cease to exist after termination. I'll update this.

clusterman/draining/queue.py
```diff
@@ -226,13 +229,14 @@ def delete_warning_messages(self, hosts: Sequence[Host]) -> None:
...
-        mesos_operator_client: Callable[..., Callable[[str], Callable[..., None]]],
+        mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
+        kube_operator_client: Optional[KubernetesClusterConnector],
     ) -> None:
         host_to_terminate = self.get_host_to_terminate()
         if host_to_terminate:
             # as for draining if it has a hostname we should down + up around the termination
-            if host_to_terminate.hostname:
-                logger.info(f'Hosts to down+terminate+up: {host_to_terminate}')
+            if host_to_terminate.scheduler in {None, 'mesos'}:
```
chlgit: If the scheduler is None, would we still need to drain it by mesos pattern?

clusterman/draining/kubernetes.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def kube_delete_node(operator_client, host):
+    operator_client.reload_state()
```
chlgit, line 26: Do we need to reload_state() again here? In other words, is _get_agent_metadata not enough?

chlgit (approved):
Looks good to me mostly. We might need to test this against norcal-devc or kubestage before it's running for production clusters.

clusterman/draining/queue.py
enochtangg: Good point, my thought was that since historically (before we had the `cluster_manger` field) we defaulted to mesos pattern. But since now we have a good mix of both, and assuming the cluster_manager is always specified, I can change this to default to terminate immediately.

clusterman/draining/kubernetes.py
enochtangg: `_get_agent_metadata()` reads directly from `self._nodes_by_ip` which is only updated when reload_state is called. I would say keep it there in case to avoid NotFound errors.

PR #28, opened 2020-03-03
Protect clusterman against malformed ec2 json tags

solarkennedy:
I'm not super sure if this should be merged. I see ec2 tags as "user generated input" and I think clusterman should protect itself against that.

At the same time, bad tags are a bug *somewhere*, is a warning in the log good enough to get that feedback?

### DIFF
--- a/clusterman/aws/aws_resource_group.py
+++ b/clusterman/aws/aws_resource_group.py
@@ -195,6 +195,11 @@ class AWSResourceGroup(ResourceGroup, metaclass=ABCMeta):
                     matching_resource_groups[rg_id] = rg
             except KeyError:
                 continue
+            except Exception as e:
+                logger.error(
+                    f'Failed to parse the identifier tags on {rg_id}, continuing anyway: {e}.\nThe tags were: {tags}'
+                )
+                continue
         return matching_resource_groups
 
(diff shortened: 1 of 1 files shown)


clusterman/aws/aws_resource_group.py
```diff
@@ -195,6 +195,9 @@ def load(cls, cluster: str, pool: str, config: Any, **kwargs: Any) -> Mapping[st
                     matching_resource_groups[rg_id] = rg
             except KeyError:
                 continue
+            except Exception as e:
+                logger.warning(f"Failed to parse the identifier tags on {rg_id}: {e}.\nThe tags were: {tags}")
```
vkhromov: It should probably be an error instead of a warning.

drmorr0:
I am uncertain how I feel about this review.  I guess I'd kinda like to know if something _is_ wrong with the tags and I'm not sure anybody's going to look at these logs.  Maybe just having clusterman crash is the right solution?  

Or another way to look at it, what can Clusterman "do" if it has bad tags?  I guess it could continue scaling the other SFRs...

Up to you -- you can ship if you want or discard if you want.

drmorr0 (approved):

PR #29, opened 2020-03-04
Auto-detect srv-configs changes with file watchers

enochtangg:
By adding some extra functionality/fixing bugs around yelp_batch (v10.1.2), this code is responsible for both adding/creating new watcher files and reloading them periodically. The system is set up to reload PoolManager and Autoscaler objects only if the watcher file has changed to reduce extra work. 

Testing Done:
* Added units to ensure re-instantiation is executed when files are updated.
* `make test`
* Manually tested on kubernetes_autoscaler, drainer, and cluster_metrics instances via `paasta local run`. Test on kubestage by:
    1. Ensure drainer was turned off on kubestage
    2. Running `clusterman manage` and decreasing capacity
    3. See on drainer logs that it was not draining/deleting nodes
    4. Push srv-config change (turning on drainer)
    5. Noticed drainer was draining nodes automatically without restarting service instance.

### DIFF
--- a/clusterman/batch/autoscaler.py
+++ b/clusterman/batch/autoscaler.py
@@ -97,7 +97,12 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
         self.logger = logger
 
         self.apps = [self.options.pool]  # TODO (CLUSTERMAN-126) someday these should not be the same thing
+        self.initialize_autoscaler()
 
+        # We don't want to watch anything here because the autoscaler bootstrap script takes care of that for us
+        self.config.watchers.clear()
+
+    def initialize_autoscaler(self):
         pool_manager = PoolManager(
             self.options.cluster,
             self.options.pool,
@@ -111,9 +116,6 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
             monitoring_enabled=(not self.options.dry_run),
             pool_manager=pool_manager,
         )
-
-        # We don't want to watch anything here because the autoscaler bootstrap script takes care of that for us
-        self.config.watchers.clear()
 
     def _get_local_log_stream(self, clog_prefix=None):
         # Overrides the yelp_batch default, which is tmp_batch_<filename> (autoscaler in this case)
@@ -135,6 +137,10 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
     def run(self):
         while self.running:
             try:
+                logger.info('Reloading watcher config files')
+                if self.reload_watchers():
+                    self.initialize_autoscaler()
+
                 self._autoscale()
             except (PoolConnectionError, EndpointConnectionError) as e:
                 logger.exception(f'Encountered a connection error: {e}')
--- a/clusterman/batch/autoscaler.py
+++ b/clusterman/batch/autoscaler.py
@@ -97,12 +97,6 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
         self.logger = logger
 
         self.apps = [self.options.pool]  # TODO (CLUSTERMAN-126) someday these should not be the same thing
-        self.initialize_autoscaler()
-
-        # We don't want to watch anything here because the autoscaler bootstrap script takes care of that for us
-        self.config.watchers.clear()
-
-    def initialize_autoscaler(self):
         pool_manager = PoolManager(
             self.options.cluster,
             self.options.pool,
@@ -116,6 +110,9 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
             monitoring_enabled=(not self.options.dry_run),
             pool_manager=pool_manager,
         )
+
+        # We don't want to watch anything here because the autoscaler bootstrap script takes care of that for us
+        self.clear_watchers()
 
     def _get_local_log_stream(self, clog_prefix=None):
         # Overrides the yelp_batch default, which is tmp_batch_<filename> (autoscaler in this case)
@@ -135,12 +132,10 @@ class AutoscalerBatch(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin)
             self.autoscaler.run(dry_run=self.options.dry_run)
 
     def run(self):
+        # self.running is a property from yelp_batch which checks version_checker if a watcher config has changed.
+        # If so, the entire batch restarts and configs for the service are reloaded.
         while self.running:
             try:
-                logger.info('Reloading watcher config files')
-                if self.reload_watchers():
-                    self.initialize_autoscaler()
-
                 self._autoscale()
             except (PoolConnectionError, EndpointConnectionError) as e:
                 logger.exception(f'Encountered a connection error: {e}')
--- a/clusterman/batch/autoscaler_bootstrap.py
+++ b/clusterman/batch/autoscaler_bootstrap.py
@@ -90,9 +90,18 @@ class AutoscalerBootstrapBatch(BatchDaemon, BatchLoggingMixin):
             self.options.pool,
             self.options.scheduler,
         )
-        self.config.watchers.append(
-            {self.options.pool: get_pool_config_path(self.options.cluster, self.options.pool, self.options.scheduler)},
-        )
+        watcher = {
+            self.options.pool:
+                get_pool_config_path(self.options.cluster,
+                                     self.options.pool,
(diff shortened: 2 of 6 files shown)


clusterman/batch/autoscaler.py
```diff
@@ -97,7 +97,12 @@ def configure_initial(self):
         self.logger = logger
 
         self.apps = [self.options.pool]  # TODO (CLUSTERMAN-126) someday these should not be the same thing
+        self.initialize_autoscaler()
 
+        # We don't want to watch anything here because the autoscaler bootstrap script takes care of that for us
+        self.config.watchers.clear()
```
kawaiwanyelp: Would this work? I recall that in `yelp_batch`, the watchers property does not deep copy, which means that you'd be clearing a copy here.

clusterman/batch/autoscaler_bootstrap.py
```diff
@@ -90,9 +90,13 @@ def configure_initial(self) -> None:
             self.options.pool,
             self.options.scheduler,
         )
-        self.config.watchers.append(
-            {self.options.pool: get_pool_config_path(self.options.cluster, self.options.pool, self.options.scheduler)},
-        )
+        watcher = {
```
kawaiwanyelp: nit: s/watcher/watcher_configs/g

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -102,15 +102,17 @@ def configure_initial(self) -> None:
             self.pools[scheduler] = get_pool_name_list(self.options.cluster, scheduler)
         for scheduler, pools in self.pools.items():
             for pool in pools:
-                self.config.watchers.append({
-                    f'{pool}.{scheduler}': get_pool_config_path(self.options.cluster, pool, scheduler),
-                })
+                watcher = {
```
kawaiwanyelp: nit: s/watcher/watcher_configs/g

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -102,15 +102,17 @@ def configure_initial(self) -> None:
...
+                    f'{pool}.{scheduler}': get_pool_config_path(self.options.cluster, pool, scheduler)
+                }
+                self.add_watcher(watcher)
                 load_cluster_pool_config(self.options.cluster, pool, scheduler, None)
+        self.initialize_clusterman_metrics_client()
 
+    def initialize_clusterman_metrics_client(self):
         self.region = staticconf.read_string('aws.region')
         self.run_interval = staticconf.read_int('batches.cluster_metrics.run_interval_seconds')
         self.logger = logger
```
kawaiwanyelp: You should probably move this line out of the function. Doesn't seem like it depends on srv-configs at all.

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -139,6 +141,10 @@ def run(self) -> None:
 
                 successful = self.write_all_metrics()
 
+                logger.info('Reloading watcher config files')
+                if self.reload_watchers():
+                    self.initialize_clusterman_metrics_client()
+
```
kawaiwanyelp: I'm wondering if you need to set watchers on the configs for and also re-initialize the pool managers here, similar to what you did for the autoscaler. (Couple of lines earlier you can see the pool managers are loaded, and have their state reloaded too.

clusterman/batch/autoscaler.py
```diff
@@ -135,6 +137,10 @@ def _autoscale(self):
     def run(self):
         while self.running:
             try:
+                logger.info('Reloading watcher config files')
+                if self.reload_watchers():
+                    self.initialize_autoscaler()
```
kawaiwanyelp: I'm wondering how expensive this is. There might be some things in here that we don't want to do again, like recreating metrics things. It might more worth it to create a function in the Autoscaler reloads everything that is dependent on srv-configs. Or better yet, instead of having the Autoscaler save a lot of srv-configs values into local variables during its init functions, just turn each of those local-variable uses into a srv-configs read (this is how it normally works).

Ditto for the PoolManagers.

kawaiwanyelp (changes requested):

clusterman/batch/autoscaler.py
enochtangg: Yeah, you're right. I think this piece of code was already here initially. I can add some functionality in yelp_batch to support this. However, I don't think it's necessary to clear these watchers at configure_initial when watchers are only loaded in during autoscaler bootstrap anyways.

clusterman/batch/cluster_metrics_collector.py
enochtangg: Could you expand on what you mean by this?

clusterman/batch/autoscaler.py
enochtangg: Most of the Autoscaler constructor actually reads from srv-configs. For example, both `ClustermanMetricsBotoClient` and `Signal` are objects that are instantiated in the Autoscaler connector. Initializing `ClustermanMetricsBotoClient` reads a `ttl_days` field and initializing Signal reads various metrics fields. I think it would it would be quite messy to just pick and choose properties to update when a lot of those properties are either inherited from `Autoscaler` or `ClusterMetricsCollector` or in an external package. In addition, even if it is an expensive operation, we aren't executing it unless something in srv-config actually changes which isn't often. Anyways, we could use more opinions on this.

clusterman/batch/cluster_metrics_collector.py
kawaiwanyelp: `self.logger = logger` doesn't use staticconf.

clusterman/batch/autoscaler.py
drmorr0: These are necessary to clear, otherwise you have both the autoscaler batch and the bootstrap batch trying to watch these files.  We just want the bootstrap batch to watch.

clusterman/batch/autoscaler.py
drmorr0: The way this used to work is that if a srv-configs file changed, yelp_batch would just set `self.running` to false and the entire batch would restart.  (see https://docs.google.com/forms/d/1McU7891ROmbh2ugtHE-9oczNDvgWyCTF59NOhZtqJUg/edit).  Is this no longer how yelp_batch works?  I haven't been following the changes too closely.  In any case, I think I would prefer that solution still to reloading srv-configs piecemeal.  There's a lot of saved state in Clusterman, and I worry about forgetting to update some of it.  Plus, it makes this code simpler.

drmorr0 (changes requested):

clusterman/batch/autoscaler.py
enochtangg: @drmorr0 You might have linked something else there. Also, how does this make the code simpler? Looking at the constructor of both PoolManager and Autoscaler, it seems like most of its attributes require reading from staticconf anyways (https://github.com/Yelp/clusterman/blob/master/clusterman/autoscaler/autoscaler.py#L84-L110). Also there are multiple child classes that inherit from these classes that would make re-initializing kinda messy. It's fair that there is a lot of saved state in clusterman and reloading the autoscaler might forget some of it. My point is that bulk of initializing the class `Autoscaler` requires reading from staticconf anyays, so why bother with pick and choosing?

clusterman/batch/autoscaler_bootstrap.py
```diff
@@ -90,9 +90,13 @@ def configure_initial(self) -> None:
...
-        self.config.watchers.append(
-            {self.options.pool: get_pool_config_path(self.options.cluster, self.options.pool, self.options.scheduler)},
-        )
+        watcher_config = {
+            self.options.pool:
+                get_pool_config_path(self.options.cluster,
+                                     self.options.pool,
+                                     self.options.scheduler)
+        }
+        self.add_watcher(watcher_config)
```
kawaiwanyelp, line 99: nit: Perhaps comment here that the watchers don't reload anything. They watch configs and they notice that anything changes, yelp_batch's version checker will restart the entire batch.

kawaiwanyelp (approved):
lgtm, just a small nit.

tests/batch/autoscaler_test.py
```diff
@@ -74,6 +74,7 @@ def test_run_ok(mock_sensu, mock_running, mock_time, mock_sleep, dry_run):
     with mock.patch('builtins.hash') as mock_hash:
         mock_hash.return_value = 0  # patch hash to ignore splaying
         batch_obj.run()
+
```
drmorr0: I don't really care whether this space is there or not, but usually if whitespace is the only change in the file I will amend the commit to make the diff smaller.  Just FYI, don't have to block on this or whatever.

drmorr0 (approved):
lgtm

PR #30, opened 2020-03-26
Make _compute_new_resource_group_targets handle resource groups with …

EvanKrall:
…capacity constraints, such as ASGs with MaxSize.

PAASTA-16471 / CLUSTERMAN-502

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -11,11 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import math
 import traceback
-from bisect import bisect
 from collections import defaultdict
 from typing import cast
 from typing import Collection
+from typing import Dict
 from typing import List
 from typing import Mapping
 from typing import MutableMapping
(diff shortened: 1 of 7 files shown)


thebostik (approved):
this LGTM. is removing the assumption of uniformity of the resource groups important to anywhere else?

drmorr0 (approved):
lgtm!  in response to @thebostik , I'm pretty sure this will be a no-op except where we want to use RIs.

clusterman/autoscaler/pool_manager.py
```diff
@@ -375,46 +376,34 @@ def _compute_new_resource_group_targets(self, new_target_capacity: float) -> Map
...
 
-        for group_to_change, new_target in zip(groups_to_change, targets_to_change):
-            targets[group_to_change.id] = new_target / coeff
+        def is_constrained(group):
+            if coeff > 0:
+                return targets[group.id] + coeff > group.max_capacity
+            else:
+                return targets[group.id] + coeff < group.min_capacity
+
+        while sum(targets.values()) * coeff < math.ceil(new_target_capacity) * coeff:
```
stug, line 391: why `math.ceil`?

stug (approved):
lg2m!

clusterman/autoscaler/pool_manager.py
EvanKrall: Otherwise it'll always scale past the new_target_capacity - if it's scaling up and new_target_capacity = 15.5, that's good, it goes up to 16; but if it's scaling down to 15.5 it would scale down to 15.

Also this made the existing tests pass.

stug: makes sense

PR #35, opened 2020-04-03
Remove the kubernetes drain and delete node code

drmorr0:
We shipped a change to ensure that nodes drain and delete themselves as
a systemd shutdown script, which is more robust (covers all failure
cases) than having it in Clusterman.  So I'm removing this code from
Clusterman so we don't have something trying to do this in two places.


clusterman/draining/queue.py
```diff
@@ -251,7 +249,6 @@ def process_termination_queue(
                 logger.info(f'Kubernetes hosts to delete k8s node and terminate: {host_to_terminate}')
                 try:
                     logger.info(f'Deleting kubernetes node')
-                    kube_delete_node(kube_operator_client, host_to_terminate)
                 except Exception as e:
                     logger.warning(f'Failed to delete {host_to_terminate.ip} node continuing to terminate anyway: {e}')
```
chlgit: I guess we can clean the log a little bit here.

chlgit (approved):

clusterman/draining/queue.py
```diff
@@ -251,7 +249,6 @@ def process_termination_queue(
                 logger.info(f'Kubernetes hosts to delete k8s node and terminate: {host_to_terminate}')
                 try:
```
enochtangg: you could probably just remove this try catch entirely

enochtangg (approved):

clusterman/draining/queue.py
drmorr0: Fixed

PR #36, opened 2020-04-09
Fix for failing trying to get unscheduleable pods in the metrics collector

drmorr0:
Clusterman was failing with

    File "/code/clusterman/kubernetes/kubernetes_cluster_connector.py", line 78, in get_unschedulable_pods
        for condition in pod.status.conditions:
    TypeError: 'NoneType' object is not iterable

This fix should catch that error and move on.


drmorr0:
Added some tests.

stug (approved):

PR #37, opened 2020-04-17
CLUSTERMAN-511 Add support for pod.affinity on metrics_collector.

gvolco:
The metrics_collector uses **pod.node_selector** to determine the pool where a pod is supposed to run. But cassandra pods are using **pod.affinity**, so I'm adding some changes to the metrics_collector to support that.

@Rob-Johnson suggested to use a similar approach to what the k8s-scheduler is doing:
- https://github.com/kubernetes/kubernetes/blob/2f08dc1b876c9c32db6a98114b25988b961d37c2/pkg/scheduler/framework/plugins/helper/node_affinity.go#L28
- https://github.com/kubernetes/kubernetes/blob/2f08dc1b876c9c32db6a98114b25988b961d37c2/pkg/apis/core/v1/helper/helpers.go#L314

I'm using some of that, but I didn't replicate the whole logic as the k8s-scheduler:
1- can accept pods with no node_selector and no affinity (which means that the pod will be scheduled anywhere)
2- If a pod has both a node_selector and affinity, the scheduler will try to satisfy both. But on our case, we will never have a pool label in both places at the same time. And we certainly would never have two different pool labels.

Tests:
- Manual run of the metrics_collector batch, having a pending cassandra pod:
```
$ python -m clusterman.batch.cluster_metrics_collector --cluster xxxx
...
INFO:__main__:Writing value 1 for metric unschedulable_pods|cluster=xxxx,pool=cassandra.kubernetes to metric store
INFO:__main__:Writing value 1.6 for metric cpus_pending|cluster=xxxx,pool=cassandra.kubernetes to metric store
INFO:__main__:Writing value 4640.0 for metric mem_pending|cluster=xxxx,pool=cassandra.kubernetes to metric store
INFO:__main__:Writing value 1792.0 for metric disk_pending|cluster=xxxx,pool=cassandra.kubernetes to metric store
INFO:__main__:Writing value 0 for metric gpus_pending|cluster=xxxx,pool=cassandra.kubernetes to metric store
...
```

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -21,6 +21,8 @@ import colorlog
 import kubernetes
 import staticconf
 from kubernetes.client.models.v1_node import V1Node as KubernetesNode
+from kubernetes.client.models.v1_node_selector_requirement import V1NodeSelectorRequirement
+from kubernetes.client.models.v1_node_selector_term import V1NodeSelectorTerm
 from kubernetes.client.models.v1_pod import V1Pod as KubernetesPod
 
 from clusterman.interfaces.cluster_connector import AgentMetadata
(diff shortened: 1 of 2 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,9 +87,34 @@ def get_unschedulable_pods(self) -> List[KubernetesPod]:
                 unschedulable_pods.append(pod)
         return unschedulable_pods
 
+    def _pod_matches_node_selector_or_affinity(self, pod: KubernetesPod) -> bool:
+        if pod.spec.node_selector:
+            for key, value in pod.spec.node_selector.items():
+                if key == self.pool_label_key:
+                    if value == self.pool:
```
Rob-Johnson: `return value == self.pool`

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,9 +87,34 @@ def get_unschedulable_pods(self) -> List[KubernetesPod]:
...
+                        return True
+                    else:
+                        return False
+
+        selector_requirement = V1NodeSelectorRequirement(
+            key=self.pool_label_key, operator='In', values=[self.pool]
+        )
+        if pod.spec.affinity and pod.spec.affinity.node_affinity:
+            node_affinity = pod.spec.affinity.node_affinity
+            if node_affinity.required_during_scheduling_ignored_during_execution:
```
Rob-Johnson, line 114: we're going to have to come back to this at some point and add support for `required_during_scheduling_required_during_execution`. I'm also wondering if we should support `preferredDuringSchedulingIgnoredDuringExecution` too?  see explanation here https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity. maybe we should just do the latter now and leave a TODO/ticket to pickup the required_during_scheduling_required_during_execution work when it's been added

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,9 +87,34 @@ def get_unschedulable_pods(self) -> List[KubernetesPod]:
                 unschedulable_pods.append(pod)
         return unschedulable_pods
 
+    def _pod_matches_node_selector_or_affinity(self, pod: KubernetesPod) -> bool:
```
Rob-Johnson, line 102: can you add a test for this somewhere to make sure it behaves as we expect?

clusterman/kubernetes/kubernetes_cluster_connector.py
gvolco: As per our last conversation, I have included support for preferredDuringSchedulingIgnoredDuringExecution.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,9 +88,49 @@ def get_unschedulable_pods(self) -> List[KubernetesPod]:
...
+        )
+
+        if pod.spec.affinity and pod.spec.affinity.node_affinity:
+            node_affinity = pod.spec.affinity.node_affinity
+            if node_affinity.required_during_scheduling_ignored_during_execution:
+                node_selector_terms = node_affinity.required_during_scheduling_ignored_during_execution.node_selector_terms  # noqa: E501
+                for selector_term in node_selector_terms:
+                    if self._selector_term_matches_requirement(selector_term, selector_requirement):
+                        return True
+            elif node_affinity.preferred_during_scheduling_ignored_during_execution:
```
Rob-Johnson: I think this needs to just be if; you can have both required_during_scheduling_ignored_during_execution and preferred_during_scheduling_ignored_during_execution

gvolco: good catch, will update the PR in a moment

gvolco: Also I'm adding a test case with both required_during_scheduling_ignored_during_execution and preferred_during_scheduling_ignored_during_execution

Rob-Johnson (approved):

PR #39, opened 2020-06-05
Account for scenarios where target_capacity is outside limit when constraining requested capacity

kawaiwanyelp:
### Description
There are some cases where an existing target_capacity will be outside the max and min capacity limits, either due to manual scaling or because the limits move. In this case, capacity requests can result in bad deltas when constraining the target capacity. This PR patches that math to account for that case.

### Testing done
`make test`

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -236,12 +236,26 @@ class PoolManager:
         """ Signals can return arbitrary values, so make sure we don't add or remove too much capacity """
 
         requested_delta = requested_target_capacity - self.target_capacity
+
+        # first, determine whether or not the delta is actually positive or negative.
+        # for example, if the current target capacity is above the maximum, the resulting delta
+        # will be negative, even if the requested delta is positive because we take the min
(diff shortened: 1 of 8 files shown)


clusterman/autoscaler/pool_manager.py
```diff
@@ -236,13 +236,25 @@ def _constrain_target_capacity(
         """ Signals can return arbitrary values, so make sure we don't add or remove too much capacity """
 
         requested_delta = requested_target_capacity - self.target_capacity
+
+        # first, determine whether or not the delta is actually positive or negative.
+        # for example, if the current target capacity is above the maximum, and the
+        # requested delta is positive, the actual delta will negative, and should be, so
```
drmorr0: can you fix this comment?  something seems off here (missing word maybe?)

drmorr0 (approved):
fix'n'ship: I think the comment here could be made a little more clear but otherwise this seems fine

clusterman/autoscaler/pool_manager.py
kawaiwanyelp: I've made it clearer.

PR #40, opened 2020-06-10
Improve information display around enabling and disabling the autoscaler

drmorr0:
### Description
1) If a cluster has autoscaling paused, `clusterman status` will print that info.
2) The `clusterman enable|disable` commands now print a response indicating success or failure.

### Testing done
manual testing on stagef
updated unit tests

### DIFF
--- a/clusterman/cli/toggle.py
+++ b/clusterman/cli/toggle.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+import time
 
 import staticconf
 
@@ -20,9 +21,10 @@ from clusterman.args import add_cluster_config_directory_arg
 from clusterman.args import add_pool_arg
 from clusterman.args import add_scheduler_arg
 from clusterman.args import subparser
-from clusterman.autoscaler.autoscaler import AUTOSCALER_PAUSED
-from clusterman.autoscaler.autoscaler import CLUSTERMAN_STATE_TABLE
 from clusterman.aws.client import dynamodb
+from clusterman.util import AUTOSCALER_PAUSED
+from clusterman.util import autoscaling_is_paused
+from clusterman.util import CLUSTERMAN_STATE_TABLE
 from clusterman.util import parse_time_string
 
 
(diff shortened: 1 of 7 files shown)


clusterman/cli/toggle.py
```diff
@@ -39,6 +42,17 @@ def disable(args: argparse.Namespace) -> None:
         Item=state,
     )
 
+    time.sleep(1)  # Give DynamoDB some time to settle
+    if not autoscaling_is_paused(args.cluster, args.pool, args.scheduler, arrow.now()):
+        print('Something went wrong!  The autoscaler is NOT paused')
+    else:
+        now_str = str(parse_time_string('now').to('local'))
+        until_str = str(parse_time_string(args.until).to('local'))
```
kawaiwanyelp: If `args.util` is None, `parse_time_string` will raise an `AttributeError` on [this](https://github.com/Yelp/clusterman/blob/master/clusterman/util.py#L176) line because it does not have a `split` function. Should probably just put it under the `args.until` conditional you have later in the function.

clusterman/cli/toggle.py
```diff
@@ -48,6 +62,11 @@ def enable(args: argparse.Namespace) -> None:
             'entity': {'S': f'{args.cluster}.{args.pool}.{args.scheduler}'},
         }
     )
+    time.sleep(1)  # Give DynamoDB some time to settle
+    if autoscaling_is_paused(args.cluster, args.pool, args.scheduler, arrow.now()):
+        print('Something went wrong!  The autoscaler is paused')
+    else:
+        print(f'The autoscaler for {args.cluster}.{args.pool}.{args.scheduler} was enabled at {arrow.now()}')
```
kawaiwanyelp: nit: I think you should use the same `arrow.now()` for both the pause check and the enable message (i.e. set it as a variable).

kawaiwanyelp (changes requested):

clusterman/cli/toggle.py
drmorr0: whoops, good catch.

clusterman/cli/toggle.py
drmorr0: fixed

kawaiwanyelp (approved):

PR #41, opened 2020-06-11
Add timestamp to autoscaler pause entry in DynamoDB

kawaiwanyelp:
### Description
This PR adds a Sensu alert that tickets if an autoscaler is paused for too long. By default, "too long" is 24 hours or 86400 seconds. 

### Testing Done
make test
dry-run manual testing against a a dynamodb entry created manually

### Notes
**update** - I've reduced this PR to adding a timestamp to the pause entry in DynamoDB. The alert will be created elsewhere, outside of Clusterman.


kawaiwanyelp:
After talking with @drmorr0, we've decided to move the check outside of Clusterman. However, we'll still need to know how long Clusterman has been paused for, so I've reduced this PR to only adding a timestamp to the pause entry we created in DynamoDB when disabling an autoscaler.

drmorr0 (approved):

PR #42, opened 2020-06-11
suppress modify_target_capacity calls when new = old

drmorr0:
### Description

When the autoscaler is running very frequently, we can be spamming AWS with `modify_target_capacity` calls even when the new target capacity hasn't changed.
Here we suppress those calls when the new target capacity = the old target capacity.

### Testing Done

Fixed the integration tests to catch this case.


gvolco (approved):
lgtm

PR #43, opened 2020-06-16
Output error msg if cluster, pool, or scheduler does not exist in CLI, cluster_metrics_collector, or drainer

kawaiwanyelp:
### Description
This PR just adds a helpful log/output whenever a cluster, pool, or scheduler does not exist. In the cluster_metrics_collector and drainer, the cluster check is most relevant. For the CLI, we now also check if the specified pool exists and is scheduled by the given scheduler.

### Testing Done
manual testing against a bad cluster, pool, and scheduler on the CLI, drainer, and cluster_metrics_collector
`make test`

### DIFF
--- a/clusterman/batch/cluster_metrics_collector.py
+++ b/clusterman/batch/cluster_metrics_collector.py
@@ -46,13 +46,13 @@ from clusterman.batch.util import suppress_request_limit_exceeded
 from clusterman.config import get_pool_config_path
 from clusterman.config import load_cluster_pool_config
 from clusterman.config import setup_config
+from clusterman.exceptions import ClusterNotFoundError
 from clusterman.mesos.metrics_generators import ClusterMetric
 from clusterman.mesos.metrics_generators import generate_framework_metadata
 from clusterman.mesos.metrics_generators import generate_kubernetes_metrics
 from clusterman.mesos.metrics_generators import generate_simple_metadata
 from clusterman.mesos.metrics_generators import generate_system_metrics
 from clusterman.util import All
-from clusterman.util import ClusterNotFoundError
 from clusterman.util import get_pool_name_list
 from clusterman.util import sensu_checkin
 from clusterman.util import setup_logging
@@ -104,7 +104,7 @@ class ClusterMetricsCollector(BatchDaemon, BatchLoggingMixin, BatchRunningSentin
                 self.pools[scheduler] = get_pool_name_list(self.options.cluster, scheduler)
             except ClusterNotFoundError as e:
                 logger.error(e)
-                raise SystemExit
+                raise
         for scheduler, pools in self.pools.items():
             for pool in pools:
                 watcher_config = {
--- a/clusterman/batch/cluster_metrics_collector.py
+++ b/clusterman/batch/cluster_metrics_collector.py
@@ -52,6 +52,7 @@ from clusterman.mesos.metrics_generators import generate_kubernetes_metrics
 from clusterman.mesos.metrics_generators import generate_simple_metadata
 from clusterman.mesos.metrics_generators import generate_system_metrics
 from clusterman.util import All
+from clusterman.util import ClusterNotFoundError
 from clusterman.util import get_pool_name_list
 from clusterman.util import sensu_checkin
 from clusterman.util import setup_logging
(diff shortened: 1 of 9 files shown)


clusterman/util.py
```diff
@@ -64,6 +64,14 @@ class ClustermanResources(NamedTuple):
     gpus: float = 0
 
 
+class ClusterNotFoundError(Exception):
```
drmorr0: Can we stick this in `clusterman/exceptions.py`?

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -99,7 +100,11 @@ def configure_initial(self) -> None:
         # first to load the cluster config path, and then read all the entries in that directory
         self.pools: MutableMapping[str, List[str]] = {}
         for scheduler in {'mesos', 'kubernetes'}:
-            self.pools[scheduler] = get_pool_name_list(self.options.cluster, scheduler)
+            try:
+                self.pools[scheduler] = get_pool_name_list(self.options.cluster, scheduler)
+            except ClusterNotFoundError as e:
```
drmorr0, line 105: How is this any different than what would happen before?  In particular it seems like we're hiding the stack trace, which is maybe valuable?  

`logger.exception` would still print the stack trace, but also the uncaught exception would print the stack trace, so I'm not sure why we need to catch it here and then raise `SystemExit`

clusterman/config.py
```diff
@@ -86,6 +89,32 @@ def load_cluster_pool_config(cluster: str, pool: str, scheduler: str, signals_br
...
+                msg = (
+                    f"Pool '{pool}' is scheduled in cluster '{cluster}' by {schedulers}, "
+                    f"not '{scheduler}'"
+                )
+            else:
+                msg = f"Pool '{pool}' does not exist in cluster '{cluster}'"
+        else:
+            msg = f"Cluster '{cluster}' does not exist"
+        print(Fore.RED + msg + Style.RESET_ALL)
+        raise SystemExit
```
drmorr0: Can we raise something other than `SystemExit` here?

clusterman/config.py
```diff
@@ -86,6 +89,32 @@ def load_cluster_pool_config(cluster: str, pool: str, scheduler: str, signals_br
...
+                schedulers = [os.path.splitext(f)[1][1:] for f in other_sched_files]
+                msg = (
+                    f"Pool '{pool}' is scheduled in cluster '{cluster}' by {schedulers}, "
+                    f"not '{scheduler}'"
+                )
+            else:
+                msg = f"Pool '{pool}' does not exist in cluster '{cluster}'"
+        else:
+            msg = f"Cluster '{cluster}' does not exist"
+        print(Fore.RED + msg + Style.RESET_ALL)
```
drmorr0: Can we log this instead of printing it?

clusterman/draining/queue.py
```diff
@@ -400,10 +401,14 @@ def terminate_host(host: Host) -> None:
...
-    for pool in get_pool_name_list(args.cluster, 'mesos'):
-        load_cluster_pool_config(args.cluster, pool, 'mesos', None)
-    for pool in get_pool_name_list(args.cluster, 'kubernetes'):
-        load_cluster_pool_config(args.cluster, pool, 'kubernetes', None)
+    try:
+        for pool in get_pool_name_list(args.cluster, 'mesos'):
+            load_cluster_pool_config(args.cluster, pool, 'mesos', None)
+        for pool in get_pool_name_list(args.cluster, 'kubernetes'):
+            load_cluster_pool_config(args.cluster, pool, 'kubernetes', None)
+    except ClusterNotFoundError as e:
```
drmorr0, line 409: Same comment as above; why not just skip the try/except block?

drmorr0 (changes requested):
A few relatively minor comments.  Thanks for doing this!

clusterman/batch/cluster_metrics_collector.py
kawaiwanyelp: I was trying to hide the stack trace yes to print a more human friendly message. I can just let it through.

clusterman/util.py
kawaiwanyelp: Yep.

clusterman/config.py
kawaiwanyelp: This function is only ever called during start up so if something fails, the process needs to fail. To make it a little cleaner, I've changed it to `sys.exit`.

clusterman/config.py
kawaiwanyelp: Sure.

clusterman/draining/queue.py
kawaiwanyelp: I've changed to just log then continue to raise.

PR #45, opened 2020-07-06
slight refactor before importing existing pending-pods signal

drmorr0:
### Description

This is a pretty straightforward refactor: it creates a new interface that signals are expected to inherit from, and it moves the existing `Signal` class to an `ExternalSignal` class.  The `Signal` interface return signature changes slightly, since eventually we will want to return a list of `ResourceRequest`s instead of a single aggregate `ResourceRequest`.

The next steps are going to be to a) import the logic from our existing pending pods signal into `pending_pods_signal.py`, and then b) add a srv-configs toggle to determine whether we are using the "external" version of the signal or the "internal" version.

### Testing Done

`make test` passes
No new features added, so no new tests or anything need to be written.

### DIFF
--- a/clusterman/interfaces/signal.py
+++ b/clusterman/interfaces/signal.py
@@ -1,0 +1,92 @@
+# Copyright 2019 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from abc import ABCMeta
+from abc import abstractmethod
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Union
+
+import arrow
+import staticconf
+from staticconf.errors import ConfigurationError
+
+from clusterman.exceptions import NoSignalConfiguredException
+from clusterman.exceptions import SignalValidationError
+
+SignalResponseDict = Dict[str, Optional[float]]
+
+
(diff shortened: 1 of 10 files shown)


clusterman/interfaces/signal.py
```diff
@@ -0,0 +1,89 @@
...
+     ) -> Union[SignalResponseDict, List[SignalResponseDict]]:
+        """ Compute a signal and return either a single response (representing an aggregate resource request), or a
+        list of responses (representing per-pod resource requests)
+
+        :param timestamp: a Unix timestamp to pass to the signal as the "current time"
+        :param retry_on_broken_pipe: if the signal socket pipe is broken, restart the signal process and try again
+        :returns: a dict of resource_name -> requested resources from the signal
+        :raises SignalConnectionError: if the signal connection fails for some reason
+        """
+        pass
```
EvanKrall, line 92: `raise NotImplementedError`? `return []`? I'm surprised this passes mypy.

EvanKrall (commented):
So the intent is for these `List[SignalResponseDict]`s to contain one SignalResponseDict per pod?

I feel like in the future we're going to want to know more about that pod, in particular any constraints it has. Any way to make that easy to slot in? `List[SomeObjectThatContainsASignalResponseDictAndHasRoomToGrow]`?

drmorr0:
> So the intent is for these `List[SignalResponseDict]`s to contain one SignalResponseDict per pod?
> 
> I feel like in the future we're going to want to know more about that pod, in particular any constraints it has. Any way to make that easy to slot in? `List[SomeObjectThatContainsASignalResponseDictAndHasRoomToGrow]`?

Yes, this is the eventual intent.  My initial goal here is just to maintain the status quo and, so I'm kinda inclined to make this change down the line a little ways.

clusterman/interfaces/signal.py
drmorr0: Whoops.  I forgot to make this an abstractmethod.  I think mypy just accepts `pass` in all cases, but making it an abstractmethod will error out if the child class doesn't implement the method.  This is the pattern we've used in the other interfaces.

EvanKrall (approved):

PR #46, opened 2020-07-09
Create the PendingPodsSignal

drmorr0:
### Description

This copies the MostRecentResourcesPending signal from the
clusterman_signals repo and into our existing codebase.  In addition, we
add support back in for `paasta_boost`, since the current
MostRecentResourcesPending signal did not use `paasta boost`.

The goal here is have the signal running in the main Clusterman process
instead of communicating over supervisord; once we have this working, we
can have the signal start returning more fine-grained information about
what needs to be scheduled and where.

### Testing Done

* New unit tests for the PendingPodsSignal
* New itests to confirm that the end-to-end PPS works correctly with and without boost

### Rollout Plan

* We need to add the right "boost" key to the required metrics for the default signal in srv-configs
* We can then roll this out slowly by just deleting the pool-defined signal block in srv-configs for each pool
* boost isn't going to work correctly until we make a change to paasta-tools to append the scheduler to the boost key

### DIFF
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -35,6 +35,7 @@ from clusterman.interfaces.signal import Signal
 from clusterman.monitoring_lib import get_monitoring_client
 from clusterman.signals.external_signal import ExternalSignal
 from clusterman.signals.external_signal import SignalResponseDict
+from clusterman.signals.pending_pods_signal import PendingPodsSignal
 from clusterman.util import autoscaling_is_paused
 from clusterman.util import ClustermanResources
 from clusterman.util import get_cluster_dimensions
@@ -96,15 +97,26 @@ class Autoscaler:
 
         self.mesos_region = staticconf.read_string('aws.region')
         self.metrics_client = metrics_client or ClustermanMetricsBotoClient(self.mesos_region)
-        self.default_signal = ExternalSignal(
-            self.cluster,
-            self.pool,
-            self.scheduler,
-            '__default__',
-            DEFAULT_NAMESPACE,
-            self.metrics_client,
-            signal_namespace=staticconf.read_string('autoscaling.default_signal_role'),
-        )
+        self.default_signal: Signal
+        if staticconf.read_bool('autoscale_signal.internal', default=False):
+            self.default_signal = PendingPodsSignal(
+                self.cluster,
+                self.pool,
+                self.scheduler,
+                '__default__',
+                DEFAULT_NAMESPACE,
+                self.metrics_client,
+            )
+        else:
+            self.default_signal = ExternalSignal(
+                self.cluster,
+                self.pool,
+                self.scheduler,
+                '__default__',
+                DEFAULT_NAMESPACE,
+                self.metrics_client,
+                signal_namespace=staticconf.read_string('autoscaling.default_signal_role'),
+            )
         self.signal = self._get_signal_for_app(self.apps[0])
         logger.info('Initialization complete')
 
--- a/clusterman/interfaces/signal.py
+++ b/clusterman/interfaces/signal.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from abc import ABCMeta
 from abc import abstractmethod
+from collections import defaultdict
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -20,17 +21,30 @@ from typing import Union
 
 import arrow
 import staticconf
-from staticconf.errors import ConfigurationError
+from clusterman_metrics import APP_METRICS
+from clusterman_metrics import ClustermanMetricsBotoClient
+from clusterman_metrics import MetricsValuesDict
+from clusterman_metrics import SYSTEM_METRICS
+from mypy_extensions import TypedDict
 
-from clusterman.exceptions import NoSignalConfiguredException
+from clusterman.exceptions import MetricsError
 from clusterman.exceptions import SignalValidationError
+from clusterman.util import get_cluster_dimensions
 
 SignalResponseDict = Dict[str, Optional[float]]
+
+
+class MetricsConfigDict(TypedDict):
+    name: str
+    type: str
+    minute_range: int
+    regex: bool
 
 
 class Signal(metaclass=ABCMeta):
     def __init__(
         self,
+        name: str,
         cluster: str,
         pool: str,
         scheduler: str,
@@ -49,11 +63,7 @@ class Signal(metaclass=ABCMeta):
         """
         reader = staticconf.NamespaceReaders(config_namespace)
 
-        try:
-            self.name: str = reader.read_string('autoscale_signal.name')
-        except ConfigurationError:
-            raise NoSignalConfiguredException(f'No signal was configured in {config_namespace}')
-
+        self.name = name
         self.cluster: str = cluster
         self.pool: str = pool
         self.scheduler: str = scheduler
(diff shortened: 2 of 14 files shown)


clusterman/autoscaler/autoscaler.py
```diff
@@ -96,14 +97,13 @@ def __init__(
 
         self.mesos_region = staticconf.read_string('aws.region')
         self.metrics_client = metrics_client or ClustermanMetricsBotoClient(self.mesos_region)
-        self.default_signal = ExternalSignal(
+        self.default_signal = PendingPodsSignal(
```
EvanKrall: is this going to require a config change for any of our mesos pools?

clusterman/interfaces/signal.py
```diff
@@ -90,3 +100,54 @@ def evaluate(
...
+    app: str,
+    metrics_client: ClustermanMetricsBotoClient,
+    required_metrics: List,
+    end_time: arrow.Arrow,
+) -> MetricsValuesDict:
+    """ Get the metrics required for a signal """
+
+    metrics: MetricsValuesDict = defaultdict(list)
+    metric_dict: MetricsConfigDict
+    for metric_dict in required_metrics:
```
EvanKrall: seems better to annotate `required_metrics` as `List[MetricsValueDict]`

clusterman/interfaces/signal.py
```diff
@@ -90,3 +100,54 @@ def evaluate(
...
+            query_results = metrics_client.get_metric_values(
+                metric_dict['name'],
+                metric_dict['type'],
+                start_time.timestamp,
+                end_time.timestamp,
+                is_regex=metric_dict['regex'],
+                extra_dimensions=dims,
+                app_identifier=app,
+            )
+            for metric_name, ts in query_results.items():
```
EvanKrall: is `ts` a timestamp or time series? I think it must be a time series but it took me a few reads of the code to figure that out. Can you make this var name longer?

clusterman/signals/pending_pods_signal.py
```diff
@@ -0,0 +1,113 @@
...
+        allocated = metrics.get(resource + '_allocated', [])
+        pending = metrics.get(resource + '_pending', [])
+        latest_allocated = max(allocated, default=(None, None))
+        latest_pending = max(pending, default=(None, None))
+
+        if latest_allocated[1] is not None and latest_pending[1] is not None:
+            resource_request[resource] = latest_allocated[1] + latest_pending[1]
+        else:
+            # This could be None but that's OK
+            resource_request[resource] = latest_allocated[1]
```
EvanKrall, line 75: Is this None if there's nothing currently running on the pool? In that case we'd probably want to scale up for pending pods.

itests/autoscaler_scaling.feature
```diff
@@ -34,12 +34,28 @@ Feature: make sure the autoscaler scales to the proper amount
...
+          And the autoscaler should scale rg1 to <rg1_target> capacity
+          And the autoscaler should scale rg2 to <rg2_target> capacity
+
+      Examples:
+        | pending   | boost | rg1_target | rg2_target |
+        | 0         |       | 10         | 10         |
+        | 19        |       | 13         | 12         |
+        | 1000      |       | 50         | 50         |
+        | 0         | 2     | 19         | 18         |
+        | 50        | 2     | 37         | 36         |
```
EvanKrall: Seems like there's some state that matters here but isn't defined within this scenario, like how 19 pending turns into 5 target capacity, why 1000 pending doesn't launch 250 capacity, and why 2x boost doesn't quite double the cluster compared to the first example. Maybe the `Given` line could include the max, current allocated cpus, and whatever else is required to calculate how much target capacity for a given number of CPUs.

clusterman/signals/external_signal.py
```diff
@@ -152,6 +148,7 @@ def _connect_to_signal_process(self) -> socket.socket:
         """
         # this creates an abstract namespace socket which is auto-cleaned on program exit
         signal_conn = socket.socket(socket.AF_UNIX)
+        print(f'{self.signal_namespace}-{self.name}-{self.app}-socket')
```
kawaiwanyelp: This looks like a debugging statement :P

tests/interfaces/signal_test.py
```diff
@@ -0,0 +1,64 @@
...
+from clusterman.interfaces.signal import get_metrics_for_signal
+
+
+@pytest.mark.parametrize('end_time', [arrow.get(3600), arrow.get(10000), arrow.get(35000)])
+def test_get_metrics(end_time):
+
+    required_metrics = staticconf.read_list(
+        'autoscale_signal.required_metrics',
+        namespace='bar.mesos_config',
+        default=[],
```
kawaiwanyelp: nit: I'd remove this so that in the event the pool config is somehow removed, we'll see an error with reading the list, instead of an error with the assert lower down.

tests/signals/pending_pods_signal_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_most_recent_values(pending_pods_signal):
+    metrics = {
+        'cpus_allocated': [(900, 250), (1000, 150)],
+        'mem_allocated': [(900, 1250), (1000, 1000)],
+        'disk_allocated': [(900, 600), (1000, 500)],
+        'gpus_allocated': [(900, 0), (1000, None)],
+        'cpus_pending': [(1000, 543)],
+    }
+    print(metrics)
```
kawaiwanyelp: ;)

itests/autoscaler_scaling.feature
```diff
@@ -34,12 +34,28 @@ Feature: make sure the autoscaler scales to the proper amount
...
          Then a ResourceRequestError is raised
 
     Scenario: the autoscaler does nothing when it is paused
-        Given an autoscaler object
+        Given a mesos autoscaler object
          When the autoscaler is paused
           And the signal resource request is 1000 cpus
          Then the autoscaler should do nothing
+
+    @wip
```
kawaiwanyelp: ;)

kawaiwanyelp (changes requested):
The logic and testing look good. Just some small nits and testing leftovers.

clusterman/signals/external_signal.py
drmorr0: Maybe cuz it is :P

clusterman/autoscaler/autoscaler.py
drmorr0: Oh, good catch.  I think it would've been OK anyways but better to put the feature toggle here too.

clusterman/signals/pending_pods_signal.py
drmorr0: No, it's `None` if there's no data.  It's `0` if there's nothing currently running on the pool.  False-y values for the win?

itests/autoscaler_scaling.feature
drmorr0: Grumble.  I was trying to remember all the math that went into this and was hoping you'd just ignore it... I'll figure it out and leave some comments for future generations.

tests/interfaces/signal_test.py
kawaiwanyelp: O, I specifically meant removing the `default=[]` line, not the entirety of what was highlighted, that's my bad.

drmorr0:
One update from the previous iteration: we no longer require the PendingPodsSignal to specify its required resources in srv-configs, that's unnecessary now and another place this can break things if we forget or get things out of sync.

kawaiwanyelp (approved):

PR #47, opened 2020-07-14
bump metrics check TTL to hopefully make this less noisy

drmorr0:
### Description

The metrics collector alert has been firing a lot in pnw-prod, this increases the timeout (everywhere, unfortunately, since it's not quite that fine-grained).

### Testing Done

n/a

### DIFF
--- a/clusterman/batch/cluster_metrics_collector.py
+++ b/clusterman/batch/cluster_metrics_collector.py
@@ -147,7 +147,7 @@ class ClusterMetricsCollector(BatchDaemon, BatchLoggingMixin, BatchRunningSentin
                         output='OK: clusterman cluster_metrics was successful',
                         check_every='1m',
                         source=self.options.cluster,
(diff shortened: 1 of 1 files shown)


mattmb (approved):
👍 20m is quite long but also I don't think it's that dangerous. if this lowers the noise to 0 then I think it's fine. If it's still paging after this I think we'll have to dig deeper or just hope that our migration to k8s speeds up the Mesos API

drmorr0:
> 👍 20m is quite long but also I don't think it's that dangerous. if this lowers the noise to 0 then I think it's fine. If it's still paging after this I think we'll have to dig deeper or just hope that our migration to k8s speeds up the Mesos API

I did a brief scan of the alerts that we've been getting, _most_ of the time they resolve within 20 minutes.  If the timeout lasts longer than that I'm going to start getting worried about autoscaling on stale data.

PR #49, opened 2020-07-15
fix handling of Decimal types from the metrics lib

drmorr0:
### Description

The metrics lib returns Decimals which need to be converted to floats or they won't play nicely with the rest of the autoscaler.  This was failing in stagef.

### Testing Done

make test -- modified the itests so they would fail in the same way, and then confirmed that the changes fix them.


kawaiwanyelp (approved):
Thanks for handling the tests too!

PR #50, opened 2020-07-21
Update the pending pods signal to use data directly from Kubernetes

drmorr0:
### Description

This code makes a few changes:

1) We read the list of pending pods directly from the Kubernetes
apiserver instead of it having to round-trip through the metrics
collector and back.  In theory this will improve Clusterman's
responsiveness to unschedulable pods.

2) We classify the reason for a pod being unschedulable.  Kubernetes
doesn't (seem to) provide this information, so we have to figure it out
ourselves.  Right now, the only two options are `InsufficientResources`
and `Unknown`, but this lays the groundwork for supporting other types
of "unschedulable" scenarios.

3) For any pods that are unschedulable because of
`InsufficientResources` we now request *double* the amount of resources
for them.  This obviously isn't the long-term solution that we want, but
hopefully in the short term it will alleviate some of the bin-packing
issues we've seen in various pools.

4) `paasta boost` will _no longer_ consider pending pods when it boosts,
just allocated resources... I'm not sure if this is the correct behavior
or not but `paasta boost` doesn't work right now anyways.  :)

5) fixed some issues with the acceptance testing pipeline

### Testing Done

`make test` passes; added new tests for the `PendingPodSignal`

### DIFF
--- a/acceptance/srv-configs/clusterman.yaml
+++ b/acceptance/srv-configs/clusterman.yaml
@@ -44,7 +44,8 @@ autoscaling:
     target_capacity_margin: 0.05
 
 autoscale_signal:
-    internal: True
+    name: ConstantSignal
+    branch_or_tag: acceptance
     period_minutes: 1
 
 sensu_config:
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -32,14 +32,15 @@ from clusterman.config import POOL_NAMESPACE
 from clusterman.exceptions import NoSignalConfiguredException
 from clusterman.exceptions import ResourceRequestError
 from clusterman.interfaces.signal import Signal
+from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
 from clusterman.monitoring_lib import get_monitoring_client
 from clusterman.signals.external_signal import ExternalSignal
-from clusterman.signals.external_signal import SignalResponseDict
 from clusterman.signals.pending_pods_signal import PendingPodsSignal
 from clusterman.util import autoscaling_is_paused
 from clusterman.util import ClustermanResources
 from clusterman.util import get_cluster_dimensions
 from clusterman.util import sensu_checkin
+from clusterman.util import SignalResourceRequest
 from clusterman.util import Status
 
 SIGNAL_LOAD_CHECK_NAME = 'signal_configuration_failed'
@@ -84,7 +85,7 @@ class Autoscaler:
         monitoring_client = get_monitoring_client()
         self.target_capacity_gauge = monitoring_client.create_gauge(TARGET_CAPACITY_GAUGE_NAME, gauge_dimensions)
         self.resource_request_gauges: Dict[str, Any] = {}
-        for resource in ('cpus', 'mem', 'disk'):
+        for resource in SignalResourceRequest._fields:
             self.resource_request_gauges[resource] = monitoring_client.create_gauge(
                 RESOURCE_GAUGE_BASE_NAME.format(resource=resource),
                 gauge_dimensions,
@@ -99,6 +100,8 @@ class Autoscaler:
         self.metrics_client = metrics_client or ClustermanMetricsBotoClient(self.mesos_region)
         self.default_signal: Signal
         if staticconf.read_bool('autoscale_signal.internal', default=False):
+            # we should never get here unless we're on Kubernetes; this assert makes mypy happy
+            assert isinstance(self.pool_manager.cluster_connector, KubernetesClusterConnector)
             self.default_signal = PendingPodsSignal(
                 self.cluster,
                 self.pool,
@@ -106,6 +109,7 @@ class Autoscaler:
                 '__default__',
                 DEFAULT_NAMESPACE,
                 self.metrics_client,
+                self.pool_manager.cluster_connector,
             )
         else:
             self.default_signal = ExternalSignal(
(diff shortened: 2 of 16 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -59,7 +62,7 @@ def reload_state(self) -> None:
         self._pods_by_ip = self._get_pods_by_ip()
 
     def get_resource_pending(self, resource_name: str) -> float:
-        return getattr(allocated_node_resources(self.get_unschedulable_pods()), resource_name)
+        return getattr(allocated_node_resources([p for p, __ in self.get_unschedulable_pods()]), resource_name)
```
kawaiwanyelp, line 65: nit: s/`__`/`_`/g

clusterman/signals/pending_pods_signal.py
```diff
@@ -43,38 +43,46 @@ def _get_required_metrics(cluster: str, pool: str, scheduler: str, minute_range:
     return required_metrics
 
 
-def _get_max_resources(*args: SignalResponseDict) -> SignalResponseDict:
+def _get_max_allocated_resource(resource: str, metrics: MetricsValuesDict) -> Optional[float]:
+    val = max(metrics.get(f'{resource}_allocated', []), default=(None, None))[1]
+    if val:
+        return float(val)
+    else:
+        return None
```
kawaiwanyelp: nit: `return float(val) if val else None`

clusterman/signals/pending_pods_signal.py
```diff
@@ -43,38 +43,46 @@ def _get_required_metrics(cluster: str, pool: str, scheduler: str, minute_range:
...
-            resource_request[resource] = None
+    resource_request = SignalResourceRequest()
+    if not metrics and not pending_pods:
+        return resource_request
+
+    pending_pods = pending_pods or []
+    if pending_pods:
+        for pod, reason in pending_pods:
+            if reason == PodUnschedulableReason.InsufficientResources:
+                resource_request += total_pod_resources(pod) * 2
```
kawaiwanyelp: nit: I'd appreciate a comment here explaining why we're doing *2, even if it's temporary.

It also might worth logging here when a pod has a reason of unknown just so we know which pods' resource requests Clusterman couldn't satisfy.

clusterman/signals/pending_pods_signal.py
```diff
@@ -126,13 +138,13 @@ def _get_boosted_metrics(self, metrics: Dict) -> Optional[Dict]:
             boost_time, boost_factor = metrics[boost_key][-1]
 
             # Get the metrics at the time of the boost
-            return {
+            return cast(MetricsValuesDict, {
```
kawaiwanyelp: nit: Not a fan of cast where I can help it. Since `MetricsValuesDict` is a `defaultdict`, you might not need to cast if you convert the dict to a `defaultdict` instead. If it's too annoying though, feel free to ignore.

tests/signals/pending_pods_signal_test.py
```diff
@@ -3,27 +3,127 @@
...
         'cpus_allocated': [(Decimal('900'), Decimal('250')), (Decimal('1000'), Decimal('150'))],
         'mem_allocated': [(Decimal('900'), Decimal('1250')), (Decimal('1000'), Decimal('1000'))],
         'disk_allocated': [(Decimal('900'), Decimal('600')), (Decimal('1000'), Decimal('500'))],
         'gpus_allocated': [(Decimal('900'), Decimal('0')), (Decimal('1000'), None)],
-        'cpus_pending': [(Decimal('1000'), Decimal('543'))],
     }
+
+
+@pytest.fixture
+def pending_pods():
```
kawaiwanyelp, line 45: I'd like to see an pod here with an unknown reason for unschedulability so that we know we are properly filtering those out.

itests/steps/autoscaler.py
```diff
@@ -223,6 +227,29 @@ def mesos_autoscaler(context):
 @behave.given('a kubernetes autoscaler object')
 def k8s_autoscaler(context):
     behave.use_fixture(autoscaler_patches, context)
+    context.mock_cluster_connector.return_value.__class__ = KubernetesClusterConnector
+    context.mock_cluster_connector.return_value.get_unschedulable_pods.return_value = (
```
kawaiwanyelp, line 231: Same here, I think there should also be an unschedulable pod that has an unknown reason.

kawaiwanyelp (changes requested):
logic looks, some minor changes needed.

clusterman/kubernetes/kubernetes_cluster_connector.py
drmorr0: So this is left over from a (probably-outdated) entry in the Yelp style guide that you should use __ for unnamed variables because _ is used for string translations... _most_ other uses in Clusterman follow this pattern, which I could be argued into changing but kinda want to leave it for now?  Unless you feel strongly.

clusterman/signals/pending_pods_signal.py
drmorr0: Huh.  Apparently I can just write `MetricsValuesDict(list, {...})`.  I didn't think that would work.

clusterman/autoscaler/autoscaler.py
```diff
@@ -71,6 +72,7 @@ def __init__(
         self.cluster = cluster
         self.pool = pool
         self.scheduler = scheduler
+        logger.info(self.scheduler)
```
kawaiwanyelp: Is this here for testing? Just logging `mesos` or `kubernetes` here is a little weird.

itests/steps/autoscaler.py
```diff
@@ -223,6 +227,48 @@ def mesos_autoscaler(context):
...
+                            name='container1',
+                            resources=V1ResourceRequirements(requests={'cpu': context.pending_cpus})
+                        ),
+                    ]),
+                ),
+                PodUnschedulableReason.InsufficientResources,
+            ),
+            (
+                V1Pod(
+                    metadata=V1ObjectMeta(name='pod1'),
```
kawaiwanyelp: nit: s/pod1/pod2/g

acceptance/srv-configs/clusterman.yaml
```diff
@@ -44,7 +44,8 @@ autoscaling:
     target_capacity_margin: 0.05
 
 autoscale_signal:
-    internal: True
+    name: ConstantSignal
+    branch_or_tag: acceptance
```
kawaiwanyelp, line 48: I think you changed this since the last review? Does this make it so the tests don't use the internal pending pods sensitive signal?

kawaiwanyelp (approved):
Just a couple of small things

clusterman/autoscaler/autoscaler.py
drmorr0: whoops, yea, just testing.

acceptance/srv-configs/clusterman.yaml
drmorr0: This means that the default signal is not using the internal pending pods signal, but the pool-configured signal still is.  The default signal is just used in case the pool signal is not present or doesn't exist.  It's necessary to change this because the Mesos-specific itests are using the same base configuration files as the Kubernetes-specific itests (but they're using different pool config files).

PR #51, opened 2020-07-23
Extremely WIP: Refactor to use resource vectors instead of weights.

EvanKrall:
### Description

This halfway-done branch switches clusterman's autoscaler from using a scalar "weight" capacity internally, to working directly with vectors of resources (cpus, mem, disk, gpus).

I augmented the ClustermanResources type so that it supports basic vector math.

The idea here is that the ResourceGroup subclasses would expose a method that lists the ClustermanResources corresponding to the instance types that can be launched. (Not yet implemented.) This would then be called in `PoolManager._compute_new_resource_group_targets` to determine how much to adjust each group's target. Once we do that calculation, `PoolManager.modify_target_capacity` would call the ResourceGroup's `modify_target_capacity` with the new ClustermanResources value, and then the ResourceGroup would launch instances as necessary.

### Testing Done

This isn't anywhere near testable. I've mostly been using mypy as guidance.


clusterman/autoscaler/pool_manager.py
```diff
@@ -72,11 +72,32 @@ def __init__(
 
         self.draining_enabled = self.pool_config.read_bool('draining_enabled', default=False)
         self.draining_client: Optional[DrainingClient] = DrainingClient(cluster) if self.draining_enabled else None
-        self.min_capacity = self.pool_config.read_int('scaling_limits.min_capacity')
-        self.max_capacity = self.pool_config.read_int('scaling_limits.max_capacity')
+        self.min_capacity = ClustermanResources(
+            cpus=self.pool_config.read_float('scaling_limits.min_capacity_cpus'),
```
drmorr0, line 72: maybe the schema should be

```
scaling_limits:
  min_capacity:
    cpus: blah
    mem: blah
...
```

clusterman/autoscaler/pool_manager.py
```diff
@@ -375,79 +418,161 @@ def _choose_nodes_to_prune(
...
+                    if all([
+                        group_with_option.all_ge(group.min_capacity),
+                        total_with_option.all_ge(self.min_capacity),
+                        total_with_option.all_ge(new_target_capacity),
+                    ]):
+                        # Only consider an option if it gets us closer to our target.
+                        if total_with_option.clamp(upper_bound=new_target_capacity).any_lt(total_target_capacity):
+                            yield (group_id, option)
+
+        def heuristic(option: Tuple[str, ClusterNodeMetadata]) -> float:
```
drmorr0, line 514: As I have been thinking about this more, I am becoming increasingly uneasy about the route that we're taking here.  This change to use the heuristic is difficult to reason about, particularly for production traffic.  Note that as written, this code is going to immediately apply to all clusters/pools (including Mesos pools, Cassandra/Flink/etc).  I'm pretty worried that this change is going to cause several DARs or have other unintended consequences that will take a significant amount of time to debug.  At the very least I think we need a feature gate so that we can roll things out more slowly, and even better would be if we can split this apart into smaller chunks that are easier to reason about.

clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -97,41 +98,47 @@ def mark_stale(self, dry_run: bool) -> None:
...
             activity has completed before initiating this one. Defaults to False,
             which is the AWS default for manual scaling activities.
         """
+
+        target_capacity = self.target_capacity_weight
         # We pretend like stale instances aren't in the ASG, but actually they are so
         # we have to double-count them in the target capacity computation
         target_capacity += self._stale_capacity
+        target_capacity += sum(self.market_weight(cnm.instance.market) for cnm in actions.to_launch)
+        target_capacity -= sum(self.market_weight(cnm.instance.market) for cnm in actions.to_terminate)
```
drmorr0, line 127: I thought we were just focusing on scale-up actions here?

clusterman/autoscaler/autoscaler.py
```diff
@@ -296,51 +277,23 @@ def _compute_target_capacity(self, resource_request: SignalResourceRequest) -> f
             )
             return current_target_capacity
 
-        # If we get here, everything is non-zero and we can use the "normal" logic to determine scaling
```
drmorr0, line 299: All the logic is quite complicated to reason about, how can we be sure that we're not introducing new bugs here by replacing it?

clusterman/aws/aws_resource_group.py
```diff
@@ -204,3 +223,40 @@ def load(cls, cluster: str, pool: str, config: Any, **kwargs: Any) -> Mapping[st
     @classmethod
     def _get_resource_group_tags(cls) -> Mapping[str, Mapping[str, str]]:  # pragma: no cover
         return {}
+
+    def cached_describe_instances(self, instance_ids: Iterable[str]) -> Dict[str, InstanceDict]:
```
drmorr0, line 227: This should probably just get called in the resource group initialization?  Since we reload resource groups on every autoscaling run, let's just store everything we need when it's initialized (see also https://github.com/Yelp/clusterman/blob/master/clusterman/aws/auto_scaling_resource_group.py#L59)

drmorr0: (I'm actually a bit confused what's going on here... is this just describing all the instances in the resource group?  Why do we need to do that?)

clusterman/interfaces/types.py
```diff
@@ -27,7 +27,6 @@ class AgentMetadata(NamedTuple):
 
 class InstanceMetadata(NamedTuple):
     market: InstanceMarket
-    weight: float
```
drmorr0, line 30: Why aren't we tracking weight anymore?

clusterman/autoscaler/autoscaler.py
```diff
@@ -83,7 +81,12 @@ def __init__(
 
         gauge_dimensions = {'cluster': cluster, 'pool': pool}
         monitoring_client = get_monitoring_client()
-        self.target_capacity_gauge = monitoring_client.create_gauge(TARGET_CAPACITY_GAUGE_NAME, gauge_dimensions)
+        self.target_capacity_gauges: Dict[str, Any] = {}
+        for resource in ('cpus', 'mem', 'disk', 'gpus'):
```
drmorr0: For these and following, can we iterate over `ClustermanResources._fields`?  We do this in a few other spots and it'll clean up the code a bit and also make this more robust in the event that we add new resource types in the future.

drmorr0 (changes requested):

clusterman/autoscaler/pool_manager.py
drmorr0: relatedly: we're going to need some intermediate step to roll out any schema change so we don't break everything that's currently running.

EvanKrall: Yeah, the dict would be cleaner, but harder to roll out. With separate parameters, we can include both in the schema/config without conflicts.

(I could also do a dict with a different name but couldn't think of a better name)

clusterman/aws/aws_resource_group.py
EvanKrall: I think I was originally intending to determine EBS volume size from describe instances. But since we currently assume 300GB for all EBS-backed instances, I only need the instance type, which I get from `describe_spot_fleet_requests`. I can make this less of a hairball.

clusterman/interfaces/types.py
EvanKrall: Originally I had swapped `weight: float` for `resources: ClustermanResources`, but then ended up removing resources from InstanceMetadata in favor of `total_resources` on AgentMetadata.

But weight is calculable from market anyway.

clusterman/autoscaler/autoscaler.py
EvanKrall: good call

PR #52, opened 2020-07-27
Implement Kubernetes versions of the acceptance tests using kind

drmorr0:
### Description

The pre-existing acceptance testing was running against a Dockerized
Mesos cluster, which is going to start causing problems as we want to
test Kubernetes-specific functionality.  This change adds in a
Kubernetes cluster via `kind`, and runs the autoscaler acceptence tests
against it.

We also fix a small issue where the metrics collector crashes if it
can't load a PoolManager for a particular pool; now it will just log the
error and continue.

### How it works

Adapted from @keymone's acceptance testing for the various operators; we use kind to create a local Kubernetes cluster, and then we apply the Clusterman deployment to the kind cluster.  The Clusterman docker image is uploaded to a local (testing) Docker registry, as described [in the kind docs](https://kind.sigs.k8s.io/docs/user/local-registry/).  We're able to use a slightly modified version of the `run_instances.py` script to launch some fake EC2 instances via `moto`, and then we just use `kubectl` to watch for the Clusterman autoscaler pod to successfully start.

There are some minor differences between the internal and external builds based on what files are present and where in Travis versus internally.

### Testing Done

`make itest`
`make itest-external`

### Note

We'll need to roll out an updated version of `kind` everywhere before this will pass internal testing pipelines.

### DIFF
--- a/acceptance/Makefile
+++ b/acceptance/Makefile
@@ -1,0 +1,183 @@
+PROJECT_NAME ?= clusterman
+WHOAMI := $(shell whoami)
+KIND_CLUSTER ?= $(WHOAMI)-$(PROJECT_NAME)
+KUBECONFIG ?= .local-$(KIND_CLUSTER).conf
+DOCKER_IMAGE ?= $(PROJECT_NAME)-dev-$(WHOAMI)
+DOCKER_PORT ?= 31234
+DOCKER_REGISTRY ?= $(KIND_CLUSTER)-reg
+DOCKER_COMPOSE := COMPOSE_PROJECT_NAME=$(PROJECT_NAME)_k8s ../.tox/acceptance/bin/docker-compose -f docker-compose-k8s.yaml
+LOCAL_ENV ?= LOCAL_ENV_$(subst -,_,$(PROJECT_NAME))
+
+.PHONY: local-env-check
+local-env-check:
+	@which kind >/dev/null || (echo 'kind must be installed in PATH' && exit 1)
+	@which kubectl >/dev/null || (echo 'kubectl must be installed in PATH' && exit 1)
+	@test "$$$(LOCAL_ENV)" = "done" || (echo 'Please run `source <(make local-env)`.' && exit 1)
+
+.PHONY: local-env
+local-env:  ## Set env vars in current shell to do local cluster development: source <(make local-env)
+	@echo "export KIND_CLUSTER=$(KIND_CLUSTER)"
+	@echo "export KUBECONFIG=$(KUBECONFIG)"
+	@echo "export $(LOCAL_ENV)=done"
+
+.PHONY: local-cluster-internal
+local-cluster-internal: local-env-check .local-cluster.yaml .local-clusterman-internal.yaml .local-cluster-up
+	@echo Done.
+
+.PHONY: local-cluster-external
+local-cluster-external: local-env-check .local-cluster.yaml .local-clusterman-external.yaml .local-cluster-up
+	@echo Done.
+
+.local-cluster-up:
+	kind create cluster --name $(KIND_CLUSTER) --config .local-cluster.yaml --verbosity 2
+	kubectl -n kube-system get configmap/coredns -o yaml | \
+		sed "s|/etc/resolv.conf|$$(awk '/nameserver/ { print $$2; exit }' /etc/resolv.conf)|" | \
+		kubectl replace -f -
+	./k8s-local-docker-registry.sh $(DOCKER_REGISTRY) $(DOCKER_PORT) $(KIND_CLUSTER)
+	docker tag $(DOCKER_IMAGE) localhost:$(DOCKER_PORT)/$(DOCKER_IMAGE)
+	docker push localhost:$(DOCKER_PORT)/$(DOCKER_IMAGE)
+	docker exec $(KIND_CLUSTER)-control-plane mkdir -p /var/lib/clusterman
+	docker cp $(KUBECONFIG) $(KIND_CLUSTER)-control-plane:/var/lib/clusterman/clusterman.conf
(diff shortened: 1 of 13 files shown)


acceptance/Makefile
```diff
@@ -0,0 +1,183 @@
...
+	echo "Checking to see if Clusterman is running..."
+	running=1; \
+	for count in 1 2 3 4 5 6 7 8 9 10; do \
+		if kubectl get pods | grep -q clusterman.*Running; then echo "Clusterman running successfully!"; running=0; break; fi; \
+		echo "Not running yet, waiting 10 seconds and trying again (check $$count of 10)..."; \
+		kubectl describe pods; \
+		kubectl logs $$(kubectl get pods | grep clusterman | cut -f1 -d' '); \
+		sleep 10; \
+	done; \
+	exit $$running
```
charleskwwan, line 64: nit: What if we exit the loop have slept for the last time, but the cluster is now running? We should probably check one more time, or change where the sleep happens.

Minor issue, feel free to ignore.

clusterman/batch/cluster_metrics_collector.py
```diff
@@ -117,8 +117,12 @@ def load_pool_managers(self) -> None:
         self.pool_managers: Mapping[str, PoolManager] = {}
         for scheduler, pools in self.pools.items():
             for pool in pools:
-                logger.info(f'Loading resource groups for {pool}.{scheduler} on {self.options.cluster}')
-                self.pool_managers[f'{pool}.{scheduler}'] = PoolManager(self.options.cluster, pool, scheduler)
+                try:
+                    logger.info(f'Loading resource groups for {pool}.{scheduler} on {self.options.cluster}')
+                    self.pool_managers[f'{pool}.{scheduler}'] = PoolManager(self.options.cluster, pool, scheduler)
+                except Exception as e:
+                    logger.exception(str(e))
```
charleskwwan: nit: Pretty sure you don't need to `str` the exception.

charleskwwan (commented):
lgtm, just some minor nits.

charleskwwan (approved):
forgot to approve

acceptance/Makefile
```diff
@@ -0,0 +1,183 @@
...
+	./k8s-local-docker-registry.sh $(DOCKER_REGISTRY) $(DOCKER_PORT) $(KIND_CLUSTER)
+	docker tag $(DOCKER_IMAGE) localhost:$(DOCKER_PORT)/$(DOCKER_IMAGE)
+	docker push localhost:$(DOCKER_PORT)/$(DOCKER_IMAGE)
+	docker exec $(KIND_CLUSTER)-control-plane mkdir -p /var/lib/clusterman
+	docker cp $(KUBECONFIG) $(KIND_CLUSTER)-control-plane:/var/lib/clusterman/clusterman.conf
+	docker exec $(KIND_CLUSTER)-control-plane chmod 644 /var/lib/clusterman/clusterman.conf
+	docker exec $(KIND_CLUSTER)-control-plane sed -i "s/\(.*server: \).*/\1https:\\/\\/127.0.0.1:6443/g" /var/lib/clusterman/clusterman.conf
+	COMPOSE_PROJECT_NAME=$(PROJECT_NAME)_k8s ../.tox/acceptance/bin/docker-compose -f docker-compose-k8s.yaml up -d moto-ec2 moto-s3 moto-dynamodb
+	sleep 10
+	docker ps
```
EvanKrall: are these load-bearing or just for debugging?

service-itest-runner
```diff
@@ -11,6 +11,11 @@ IMAGE_NAME="clusterman-dev-$(whoami)"
 docker inspect --type=image "${IMAGE_NAME}" > /dev/null || \
     IMAGE_NAME="docker-paasta.yelpcorp.com:443/services-clusterman:paasta-$(git rev-parse HEAD)"
 DISTRIB_CODENAME=$(docker run -t "${IMAGE_NAME}" lsb_release -cs | tr -d '\n\r')
+if [ "${DISTRIB_CODENAME}" = "xenial" ]; then
+    CIDR_BLOCK="10.0.0.0/24"
+else
+    CIDR_BLOCK="11.0.0.0/24"
+fi
```
EvanKrall, line 20: I know this is just ported from run_instance.py, but what's the deal with this? why do we use different IP ranges for different OSes?

acceptance/Makefile
```diff
@@ -0,0 +1,183 @@
...
+	ACCEPTANCE_ROOT=. DISTRIB_CODENAME=bionic python run_instance.py \
+		$$(docker ps | grep clusterman_k8s.*ec2 2>&1 | sed -e 's|^.*:\(.*\)->5000.*$$|http://127.0.0.1:\1|') \
+		$$(docker ps | grep clusterman_k8s.*s3 2>&1 | sed -e 's|^.*:\(.*\)->5000.*$$|http://127.0.0.1:\1|') \
+		$$(docker ps | grep clusterman_k8s.*dynamodb 2>&1 | sed -e 's|^.*:\(.*\)->5000.*$$|http://127.0.0.1:\1|') \
```
EvanKrall: should probably be using `$(PROJECT_NAME)_k8s`, not `clusterman_k8s` here.

Is there no easy way to get this from `docker-compose`? Maybe parse `docker-compose ps` instead of `docker ps`?

EvanKrall (approved):

acceptance/Makefile
drmorr0: the first is load-bearing, left a comment.

service-itest-runner
drmorr0: When they run in parallel on Jenkins, they'll often end up on the same box, and then they need different IP ranges.  I guess I can leave a comment here.

PR #53, opened 2020-08-03
don't use external PyPI internally

drmorr0:
### Description

Make sure we're not using the external PyPI for internal builds.

Also, fix the internal build: In order to run the itests you need the environment variables set, but there's no way to run `source <(make local-env)` on Jenkins without creating a custom Jenkinsfile, which I didn't want to do. So instead I just export the two variables right before the appropriate Make target, which should allow the build to pass internally.

### Testing Done

I'll double-check the Jenkins logs after I push this.

### DIFF
--- a/acceptance/Makefile
+++ b/acceptance/Makefile
@@ -6,26 +6,13 @@ DOCKER_IMAGE ?= $(PROJECT_NAME)-dev-$(WHOAMI)
 DOCKER_PORT ?= 31234
 DOCKER_REGISTRY ?= $(KIND_CLUSTER)-reg
 DOCKER_COMPOSE := COMPOSE_PROJECT_NAME=$(PROJECT_NAME)_k8s ../.tox/acceptance/bin/docker-compose -f docker-compose-k8s.yaml
-LOCAL_ENV ?= LOCAL_ENV_$(subst -,_,$(PROJECT_NAME))
-
-.PHONY: local-env-check
-local-env-check:
-	@which kind >/dev/null || (echo 'kind must be installed in PATH' && exit 1)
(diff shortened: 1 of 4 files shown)


smrojas:
doesn't seem like travis should be failing

acceptance/Makefile
```diff
@@ -6,26 +6,13 @@ DOCKER_IMAGE ?= $(PROJECT_NAME)-dev-$(WHOAMI)
...
-local-env-check:
-	@which kind >/dev/null || (echo 'kind must be installed in PATH' && exit 1)
-	@which kubectl >/dev/null || (echo 'kubectl must be installed in PATH' && exit 1)
-	@test "$$$(LOCAL_ENV)" = "done" || (echo 'Please run `source <(make local-env)`.' && exit 1)
-
-.PHONY: local-env
-local-env:  ## Set env vars in current shell to do local cluster development: source <(make local-env)
-	@echo "export KIND_CLUSTER=$(KIND_CLUSTER)"
-	@echo "export KUBECONFIG=$(KUBECONFIG)"
-	@echo "export $(LOCAL_ENV)=done"
```
kawaiwanyelp, line 21: Why are you removing this? I thought this was necessary for running itests on Kubernetes?

kawaiwanyelp (approved):
Just one question about the acceptance test env stuff.

acceptance/Makefile
drmorr0: In order to run the itests you need the environment variables set, but there's no way to run `source <(make local-env)` on Jenkins without creating a custom Jenkinsfile, which I didn't want to do.  So instead I just export the two variables right before the appropriate Make target, which should allow the build to pass internally.

In hindsight I suppose I could have added that to the PR description.

smrojas (approved):
lgtm

PR #55, opened 2020-08-13
Slight refactor to ASG resource groups; also now read launch templates

drmorr0:
### Description

This change updates the AutoScalingResourceGroup to read launch
templates (the new way) instead of launch configurations (the old way).
The launch configuration code actually was not used anywhere anyways, so
it's safe to delete.

As an aside, I decided to slightly refactor the ASG resource group; I've
wanted to get rid of the timed_cached_property junk, and the resource
groups are the last place where we still had them.  Since the resource
groups get reloaded after every autoscaling run anyways, it's safe to
just put all the stuff we need cached into the __init__ method, so I do
that here.

I fixed most of the tests except for the "asg stale" itests; marking
ASGs stale doesn't actually work right now anyways, and we have a
different ticket to fix those, so I just want to skip the tests for now.

Lastly, there's a longstanding issue with moto where it didn't propogate
SFR tags correctly that has since been fixed so I decided to clean that
up in our tests.

Yes, I know the moto import in our requirements.txt is sketchy.  The
repo I'm pulling from is one of the core moto devs, and they literally
just opened a PR to test for LaunchTemplates a few hours ago.  Hopefully
this will get merged soon and we can go back to using the pre-built
package.

(see https://github.com/spulec/moto/pull/3236)

### Testing Done

`make test` passes

### DIFF
--- a/clusterman/aws/auto_scaling_resource_group.py
+++ b/clusterman/aws/auto_scaling_resource_group.py
@@ -12,36 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import pprint
-from typing import Any
-from typing import Dict
 from typing import Iterator
+from typing import List
 from typing import Mapping
 from typing import Sequence
+from typing import Tuple
 
 import colorlog
-from cached_property import timed_cached_property
-from mypy_extensions import TypedDict
-from retry import retry
 
-from clusterman.aws import CACHE_TTL_SECONDS
 from clusterman.aws.aws_resource_group import AWSResourceGroup
 from clusterman.aws.client import autoscaling
 from clusterman.aws.client import ec2
 from clusterman.aws.markets import InstanceMarket
+from clusterman.aws.response_types import AutoScalingGroupConfig
+from clusterman.aws.response_types import InstanceOverrideConfig
+from clusterman.aws.response_types import LaunchTemplateConfig
 from clusterman.util import ClustermanResources
 
 _BATCH_MODIFY_SIZE = 200
 CLUSTERMAN_STALE_TAG = 'clusterman:is_stale'
 
 logger = colorlog.getLogger(__name__)
-
-
-AutoScalingResourceGroupConfig = TypedDict(
-    'AutoScalingResourceGroupConfig',
-    {
-        'tag': str,
-    }
-)
 
 
 class AutoScalingResourceGroup(AWSResourceGroup):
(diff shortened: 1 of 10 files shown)


clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -36,12 +35,10 @@
...
-AutoScalingResourceGroupConfig = TypedDict(
-    'AutoScalingResourceGroupConfig',
-    {
-        'tag': str,
-    }
-)
+_BATCH_MODIFY_SIZE = 200
+CLUSTERMAN_STALE_TAG = 'clusterman:is_stale'
+
+logger = colorlog.getLogger(__name__)
```
kawaiwanyelp: It looks like you're double defining these? (the original ones are L32-35)

stug: maybe github is just messing up its diff, but it looks like these already exist a couple lines up?

clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -174,16 +148,41 @@ def modify_target_capacity(
...
+            launch_template_name = template['LaunchTemplateName']
+            launch_template_version = template['Version']
+            overrides: List[InstanceOverrideConfig] = []
+        except KeyError:
+            policy = self._group_config['MixedInstancesPolicy']
+            template = policy['LaunchTemplate']['LaunchTemplateSpecification']
+            launch_template_name = template['LaunchTemplateName']
+            launch_template_version = template['Version']
+            overrides = policy['Overrides']
+
```
stug: Nit: i think the `try` is kind of overbroad here.  It looks like you are only expecting a KeyError on getting `LaunchTemplate`, so maybe you should just put that in the try/except?

stug (approved):
I didn't really look at the tests, but the code changes make rough sense to me

clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -174,16 +148,41 @@ def modify_target_capacity(
 
         autoscaling.set_desired_capacity(**kwargs)
 
-    @property
-    def min_capacity(self) -> int:
-        return self._group_config['MinSize']
+    def scale_up_options(self) -> Iterator[ClustermanResources]:
+        pass
```
kawaiwanyelp: Why is this not also `NotImplementedError`?

I'm pretty sure this also fails mypy.

tox.ini
```diff
@@ -12,11 +12,11 @@ deps =
     -rrequirements.txt
     -rrequirements-dev.txt
 commands =
-    check-requirements -v
+    # check-requirements -v
```
kawaiwanyelp, line 16: nit: Please comment here that this is commented out until moto is fixed.

kawaiwanyelp (approved):

clusterman/aws/auto_scaling_resource_group.py
drmorr0: Huh.  I wonder how that happened.

clusterman/aws/auto_scaling_resource_group.py
drmorr0: fixed.

clusterman/aws/auto_scaling_resource_group.py
drmorr0: It does not fail mypy, I kinda don't care whether it's `pass` or `NotImplementedError` because that's the thing I'm putting my next PR up for.

drmorr0:
I don't know why one of the Travis builds is consistently failing but I'm gonna merge this.

PR #56, opened 2020-08-14
fix AWS schema error and log when we can't load a resource group

drmorr0:
### Description

I misread the AWS schema, and we couldn't tell why things were failing because we had a `try: ... except: continue` block.

### Testing Done

`make test`
manual testing on infrastage
we can't ship this until https://github.com/Yelp/terraform-code/pull/2310 is shipped.

### DIFF
--- a/clusterman/aws/aws_resource_group.py
+++ b/clusterman/aws/aws_resource_group.py
@@ -189,11 +189,15 @@ class AWSResourceGroup(ResourceGroup, metaclass=ABCMeta):
 
         for rg_id, tags in resource_group_tags.items():
             try:
-                identifier_tags = json.loads(tags[identifier_tag_label])
-                if identifier_tags['pool'] == pool and identifier_tags['paasta_cluster'] == cluster:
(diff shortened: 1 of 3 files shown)


clusterman/aws/aws_resource_group.py
```diff
@@ -194,6 +194,7 @@ def load(cls, cluster: str, pool: str, config: Any, **kwargs: Any) -> Mapping[st
                     rg = cls(rg_id, **kwargs)
                     matching_resource_groups[rg_id] = rg
             except KeyError:
+                logger.exception(f'Could not load resource group {rg_id}; skipping...')
```
kawaiwanyelp, line 200: For greater clarity, I would also print the exception, just in case the key we don't find isn't the one we expect.

kawaiwanyelp (approved):

clusterman/aws/aws_resource_group.py
drmorr0: `logger.exception` prints the whole traceback, including the exception.

PR #57, opened 2020-08-14
handle ASGs that are still using LaunchConfigurations

drmorr0:
### Description

Some of our ASGs aren't using `LaunchTemplates` yet.  They won't be able to use fancy features in the future.

### Testing Done

`make test`, added some new tests.
manual testing in stagef, where this failed last time

### Other

Here's the `git diff` from the last time I shipped this: https://fluffy.yelpcorp.com/i/rwgv8r3H8LhjdlgkZ65V1tPLMK6tVRJF.html

### DIFF
--- a/clusterman/aws/auto_scaling_resource_group.py
+++ b/clusterman/aws/auto_scaling_resource_group.py
@@ -15,6 +15,7 @@ import pprint
 from typing import Iterator
 from typing import List
 from typing import Mapping
+from typing import Optional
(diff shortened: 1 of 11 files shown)


kawaiwanyelp (approved):
lgtm. Looks like the previous review, just with the extra case accounting for `LaunchConfigurations`. It's a shipit from me assuming there isn't anything else changed that I missed by accident.

PR #60, opened 2020-08-20
Implement checks to see if a pod can be scheduled on a node

drmorr0:
### Description

This commit checks the following conditions to see if a pod can be
scheduled on a particular node:

0) Is the node "schedulable" from Kubernetes' perspective?
1) Is the node's "max_tasks" (i.e., the maxpods kubelet setting)
   respected?
2) Does the node match the pod's nodeSelector(s)?
3) Does the pod have the appropriate tolerations for the node's taints?

If the answer to these three questions is "yes", the node will be added
as an option for the pod to be scheduled on.

### Testing Done

`make test` passes; new tests added for `_filter_scale_up_options_for_pod` and `is_node_tolerable`

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -35,12 +35,14 @@ from clusterman.aws.util import RESOURCE_GROUPS
 from clusterman.config import POOL_NAMESPACE
 from clusterman.draining.queue import DrainingClient
 from clusterman.exceptions import AllResourceGroupsAreStaleError
+from clusterman.exceptions import ClustermanException
 from clusterman.exceptions import PoolManagerError
 from clusterman.exceptions import ResourceGroupError
 from clusterman.interfaces.cluster_connector import ClusterConnector
 from clusterman.interfaces.resource_group import ResourceGroup
 from clusterman.interfaces.types import AgentState
 from clusterman.interfaces.types import ClusterNodeMetadata
+from clusterman.kubernetes.util import is_node_tolerable
 from clusterman.kubernetes.util import total_pod_resources
 from clusterman.monitoring_lib import get_monitoring_client
 from clusterman.util import read_int_or_inf
@@ -49,6 +51,11 @@ AWS_RUNNING_STATES = ('running',)
 MIN_CAPACITY_PER_GROUP = 1
 SFX_RESOURCE_GROUP_MODIFICATION_FAILED_NAME = 'clusterman.resource_group_modification_failed'
 logger = colorlog.getLogger(__name__)
+
+
+# Just using this for flow control because exceptions aren't exceptional in Python :(
+class _FilterException(ClustermanException):
+    pass
 
 
 class PoolManager:
@@ -206,27 +213,6 @@ class PoolManager:
             for instance_metadata in group.get_instance_metadatas(state_filter)
         ]
 
-    def _filter_scale_up_options_for_pod(
-        self,
-        pod: KubernetesPod,
-        scale_up_options: Mapping[str, List[ClusterNodeMetadata]],
-    ) -> Mapping[str, List[ClusterNodeMetadata]]:
-        filtered_options: Mapping[str, List[ClusterNodeMetadata]] = defaultdict(list)
-        for group_id, options in scale_up_options.items():
-            for option in options:
-                reason = ''
-                if total_pod_resources(pod) > option.agent.allocated_resources:
-                    reason = 'insufficient resources'
-
-                if reason:
-                    logger.debug(
-                        'Skipping option {option.instance.market} for {pod.metadata.name}: {reason}'
-                    )
-                    continue
-                filtered_options[group_id].append(option)
-
-        return filtered_options
-
     def _reload_resource_groups(self) -> None:
         resource_groups: MutableMapping[str, ResourceGroup] = {}
         for resource_group_conf in self.pool_config.read_list('resource_groups'):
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -211,3 +224,150 @@ def block_meteorite_emission():
 @pytest.fixture
 def fn():
     return PiecewiseConstantFunction(1)
+
+
+@pytest.fixture
+def pod1():
```
kawaiwanyelp, line 230: These fixture names are a little non-descriptive. Descriptive names like `simple_pod` or `evictable_pod` are best, but if they are too long, then perhaps documenting the special characteristics of each pod in comments would help for people trying to follow along the tests down the line.

Especially since you've moved this to `conftest.py`, there will potentially be a lot of random `pod1`s all over the place, which can make it confusing since the implementation is not in the file it is used in.

clusterman/autoscaler/pool_manager.py
```diff
@@ -247,6 +233,44 @@ def _reload_resource_groups(self) -> None:
...
+
+                    if option.agent.max_tasks and option.agent.task_count >= option.agent.max_tasks:
+                        raise _FilterException('node running max tasks')
+
+                    for key, value in (pod.spec.node_selector or {}).items():
+                        if option.agent.labels.get(key) != value:
+                            raise _FilterException('mismatched node selector')
+
+                    if not is_node_tolerable(pod, option.agent.taints):
+                        raise _FilterException('node has un-matched taints')
```
kawaiwanyelp, line 259: I think you could go a step further and return the names of unmatched taints from `is_node_tolerable`. That way, when you raise an exception over unmatched taints, you can output them for easier debugging.

clusterman/autoscaler/pool_manager.py
```diff
@@ -247,6 +233,44 @@ def _reload_resource_groups(self) -> None:
...
+
+                    if total_pod_resources(pod) > option.agent.allocated_resources:
+                        raise _FilterException('insufficient resources')
+
+                    if option.agent.max_tasks and option.agent.task_count >= option.agent.max_tasks:
+                        raise _FilterException('node running max tasks')
+
+                    for key, value in (pod.spec.node_selector or {}).items():
+                        if option.agent.labels.get(key) != value:
+                            raise _FilterException('mismatched node selector')
```
kawaiwanyelp, line 256: I think you should do a set diff between the node selectors and agent labels to  get all labels that aren't on the node, and then output that for easier debugging.

clusterman/autoscaler/pool_manager.py
```diff
@@ -247,6 +233,44 @@ def _reload_resource_groups(self) -> None:
...
+    ) -> Mapping[str, List[ClusterNodeMetadata]]:
+        filtered_options: Mapping[str, List[ClusterNodeMetadata]] = defaultdict(list)
+        for group_id, options in scale_up_options.items():
+            for option in options:
+                try:
+                    if option.agent.unschedulable:
+                        raise _FilterException('node is not schedulable')
+
+                    if total_pod_resources(pod) > option.agent.allocated_resources:
+                        raise _FilterException('insufficient resources')
```
kawaiwanyelp, line 249: nit: I'm wondering whether or not you could also report a comparison between the pod's resources and node's resources for easier debugging.

tests/autoscaler/pool_manager_test.py
```diff
@@ -239,6 +242,93 @@ def test_successful(self, mock_logger, mock_pool_manager):
...
+    def test_filter_scale_up_options_node_not_tolerable(self, mock_pool_manager, pod1, caplog):
+        options = {
+            'rg1': [ClusterNodeMetadata(
+                agent=AgentMetadata(
+                    allocated_resources=ClustermanResources(cpus=20, mem=1000, disk=1000, gpus=0),
+                ),
+                instance=mock.Mock(),
+            )]
+        }
+        with mock.patch('clusterman.autoscaler.pool_manager.is_node_tolerable', return_value=False):
```
kawaiwanyelp, line 327: nit: I'd prefer if this were a function decorator.

kawaiwanyelp (changes requested):

clusterman/autoscaler/pool_manager.py
```diff
@@ -247,6 +233,44 @@ def _reload_resource_groups(self) -> None:
...
+        scale_up_options: Mapping[str, List[ClusterNodeMetadata]],
+    ) -> Mapping[str, List[ClusterNodeMetadata]]:
+        filtered_options: Mapping[str, List[ClusterNodeMetadata]] = defaultdict(list)
+        for group_id, options in scale_up_options.items():
+            for option in options:
+                try:
+                    if option.agent.unschedulable:
+                        raise _FilterException('node is not schedulable')
+
+                    if total_pod_resources(pod) > option.agent.allocated_resources:
```
EvanKrall, line 248: Based on my understanding of what `allocated_resources` means (the sum of the resource requests in all the pod specs on that node), I think this should probably be comparing to `option.agent.total_resources - option.agent.allocated_resources` (or make an `available_resources` property on AgentMetadata that does this calculation for you.)

tests/conftest.py
EvanKrall: +1, though I recognize that these are just being moved from one file to another.

Honestly, conftest.py bothers me a bit -- generally speaking in python you can always tell where a particular variable comes from by searching for it within the same file (it's either defined or imported somewhere in the file). Pytest fixtures usually follow that pattern too, unless they're defined in conftest.py. Then you just have to know enough about pytest to remember to look there. (Or `git grep` the codebase for the name, in which case descriptive names are helpful.)

The conftest ship has sailed in this codebase, though, and I don't see a great option for avoiding it here while still being able to share the fixture among multiple test files.

EvanKrall (approved):
fix the allocated_resources bug and ship.

tests/autoscaler/pool_manager_test.py
drmorr0: I dislike using mock.patch as a function decorator and prefer the context manager, I'm going to discard unless you feel strongly :)

tests/conftest.py
drmorr0: I cheated on this a little bit.  Instead of renaming everything I just moved the `pod1` definition to conftest and renamed it `simple_pod`, and left everything else alone.  I agree that these should maybe have better names but I think it's actually easier to just leave them alone for now, and handle them on a case-by-case basis in the future.

PR #27, opened 2020-02-25
clusterman status/manage/etc for prod clusters don't work from devboxes

mzq592:
Tested locally.
```
(virtualenv_run) mingqiz@dev54-uswest1adevc:~/clusterman (u/mingqiz/CLUSTERMAN-333) $ python3.7 -m clusterman.run status --cluster norcal-prod
CRITICAL:clusterman.mesos.util:Mesos is unreachable:

Command timed out. Please ssh to 'adhoc-prod' if you set --cluster to prod.
Querying Mesos URL: http://mesos.paasta-norcal-prod.yelp:5050/redirect`
```

PR #61, opened 2020-08-31
add helpful messages if you're running commands in the wrong place

drmorr0:
### Description

This change lets you know if you're running from dev and trying to access a prod cluster, or running a Kubernetes command not on a Kubernetes master, instead of just spitting out tracebacks.  Modified slightly from #27 .

### Testing Done

`make test`
manual testing

### DIFF
--- a/clusterman/cli/util.py
+++ b/clusterman/cli/util.py
@@ -1,0 +1,25 @@
+import argparse
+import signal
+import socket
+
+import colorlog
+
+logger = colorlog.getLogger(__name__)
+TIMEOUT_TIME_SECONDS = 10
+
+
+def timeout_wrapper(main):
+    def wrapper(args: argparse.Namespace):
+
+        # After 10s, prints a warning message if the command is running from the wrong place
+        def timeout_handler(signum, frame):
+            warning_string = 'This command is taking a long time to run; are you running from the right account?'
+            if 'yelpcorp' in socket.getfqdn():
+                warning_string += '\nHINT: try ssh\'ing to adhoc-prod or another box in the right account.'
+            logger.warning(warning_string)
+
+        signal.signal(signal.SIGALRM, timeout_handler)
+        signal.alarm(TIMEOUT_TIME_SECONDS)
+
+        main(args)
+    return wrapper
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import socket
 from collections import defaultdict
 from distutils.util import strtobool
 from typing import List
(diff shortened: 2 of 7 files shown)


askedrelic:
🎆

clusterman/cli/util.py
```diff
@@ -0,0 +1,23 @@
+import argparse
+import signal
+
+import colorlog
+
+logger = colorlog.getLogger(__name__)
+
+
+def timeout_wrapper(main):
```
kawaiwanyelp, line 11: nit: The naming specifically assumes a main function that accepts an args namespace. I think you can just make the `timeout_wrapper` accept any function, and `wrapper` accept any and all arguments. It'll just pass those to the function.

clusterman/cli/util.py
```diff
@@ -0,0 +1,23 @@
...
+    def wrapper(args: argparse.Namespace):
+
+        # After 10s, prints a warning message if the command is running from the wrong place
+        def timeout_handler(signum, frame):
+            logger.warning(
+                'This command is taking a long time to run; are you running from the right account?\n'
+                'HINT: try ssh\'ing to adhoc-prod or another box in the right account.'
```
kawaiwanyelp: nit: Since Clusterman is a open-source, we should consider using as few Yelp-specific refs as possible. Here, I think we shouldn't refer to `adhoc-prod` at all. Maybe instead specify an example ("e.g. If you're running this command for a prod cluster but you're in a dev environment, you need to SSH to a prod box").

clusterman/cli/util.py
```diff
@@ -0,0 +1,23 @@
...
+            )
+
+        signal.signal(signal.SIGALRM, timeout_handler)
+        signal.alarm(10)
```
kawaiwanyelp: For posterity, I think it's better if you made the timeout duration configurable. You could also go the extra step of making the signal configurable, but I think `SIGALRM` is less likely to change.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -55,7 +55,16 @@ def __init__(self, cluster: str, pool: Optional[str]) -> None:
 
     def reload_state(self) -> None:
         logger.info('Reloading nodes')
-        kubernetes.config.load_kube_config(staticconf.read_string(f'{self.kubeconfig_path}'))
+        try:
+            kubernetes.config.load_kube_config(staticconf.read_string(f'{self.kubeconfig_path}'))
+        except TypeError:
+            error_msg = (
+                'Could not load KUBECONFIG; is this running on Kubernetes master?\n'
+                'Hint: if this is from the CLI, try using the clusterman-k8s-<clustername> wrapper script!'
```
kawaiwanyelp: nit: Same comment here on Yelp-isms. For Yelp users, I'd suggest running an `os` command to check whether or not the wrapper exists before append this hint.

kawaiwanyelp (approved):
Small nits

clusterman/cli/util.py
drmorr0: I kinda disagree here... I think this function should be explicit that it's wrapping the `main` function of a CLI subcommand.  If it changes to be more generally-used in the future, we can update the terminology then.

clusterman/cli/util.py
drmorr0: I agree that it should not be a magic number but I don't think exposing it as a config option makes sense.  I'll make that change.

clusterman/kubernetes/kubernetes_cluster_connector.py
drmorr0: fixed.

clusterman/cli/util.py
drmorr0: fixed.

PR #62, opened 2020-09-10
`clusterman status --json`

drmorr0:
### Description

Clusterman status can now produce JSON output for ease of parsing by
machines or by other scripts.

I also deleted some Mesos-specific terms where the code is more general.

### Testing Done

`make test` passes
manual testing: 
  - JSON output looks like [this](https://fluffy.yelpcorp.com/i/C1XlB7dszDxjG5kS9vHMXdJGdcwvM1nn.html)
  - Human-readable output looks like [this](https://fluffy.yelpcorp.com/i/cl65xxDRh8MnWTV7cMv243jKCLWWPP6V.html)

### DIFF
--- a/clusterman/cli/status.py
+++ b/clusterman/cli/status.py
@@ -13,83 +13,173 @@
 # limitations under the License.
 import argparse
 import sys
+from typing import Iterable
+from typing import List
+from typing import Mapping
+from typing import Optional
+from typing import Sequence
 
 import arrow
+import simplejson as json
 from colorama import Fore
 from colorama import Style
 from humanfriendly import format_size
 from humanfriendly import format_timespan
+from mypy_extensions import TypedDict
 
 from clusterman.args import add_cluster_arg
 from clusterman.args import add_cluster_config_directory_arg
+from clusterman.args import add_json_arg
 from clusterman.args import add_pool_arg
 from clusterman.args import add_scheduler_arg
 from clusterman.args import subparser
 from clusterman.autoscaler.pool_manager import PoolManager
+from clusterman.aws.markets import InstanceMarket
 from clusterman.cli.util import timeout_wrapper
+from clusterman.interfaces.resource_group import ResourceGroup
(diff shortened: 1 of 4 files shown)


clusterman/cli/status.py
```diff
@@ -109,7 +114,43 @@ def _write_summary(manager: PoolManager) -> None:
...
+    result = {
+        'disabled': autoscaling_is_paused(manager.cluster, manager.pool, manager.scheduler, arrow.now()),
+        'target_capacity': manager.target_capacity,
+        'fulfilled_capacity': manager.fulfilled_capacity,
+        'non_orphan_fulfilled_capacity': manager.non_orphan_fulfilled_capacity,
+        'resource_groups': [
+            _json_agent_details(metadata)
+            for metadata in node_metadatas
+            for group in manager.resource_groups.values()
+            if metadata.instance.group_id != group.id
```
stug: I'm not sure I follow this -- why is it `!=` ?  In the non-json version you skip printing if the group ids don't match.

clusterman/cli/status.py
```diff
@@ -109,7 +114,43 @@ def _write_summary(manager: PoolManager) -> None:
...
+        ],
+    }
+    print(json.dumps(result))
+
+
+def print_status(manager: PoolManager, args: argparse.Namespace) -> None:
```
stug, line 206: Wouldn't it be better to reimplent this to just parse the dictionary created by `print_status_json`?  Otherwise I worry the two implementations will drift.

stug (changes requested):

clusterman/cli/status.py
drmorr0: oops, you're right.  good catch.

clusterman/cli/status.py
drmorr0: Yea, fair.  OK.

stug (approved):

PR #67, opened 2020-10-08
Consider Pending pods when calculating resource utilization

JoeMalt:
### Description

Details in [CLUSTERMAN-561](https://jira.yelpcorp.com/browse/CLUSTERMAN-561)

In short, we don't consider Pending pods when looking at what's on a node, but k8s does for scheduling purposes. This means if lots of pods are stuck Pending (e.g. `ImagePullBackOff`), k8s might be out of capacity but Clusterman doesn't know it.

### Testing Done

`make test` passes, but this method doesn't have any test coverage and I don't think such a small change merits writing a bunch of new tests.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -217,10 +217,11 @@ class KubernetesClusterConnector(ClusterConnector):
         }
 
     def _get_pods_by_ip(self) -> Mapping[str, List[KubernetesPod]]:
+        KUBERNETES_SCHEDULED_PHASES = {'Pending', 'Running'}
         all_pods = self._core_api.list_pod_for_all_namespaces().items
         pods_by_ip: Mapping[str, List[KubernetesPod]] = defaultdict(list)
         for pod in all_pods:
-            if (pod.status.phase == 'Running' or pod.status.phase == 'Pending') \
+            if pod.status.phase in KUBERNETES_SCHEDULED_PHASES \
                     and pod.status.host_ip in self._nodes_by_ip:
                 pods_by_ip[pod.status.host_ip].append(pod)
         return pods_by_ip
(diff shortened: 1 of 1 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -220,7 +220,8 @@ def _get_pods_by_ip(self) -> Mapping[str, List[KubernetesPod]]:
         all_pods = self._core_api.list_pod_for_all_namespaces().items
         pods_by_ip: Mapping[str, List[KubernetesPod]] = defaultdict(list)
         for pod in all_pods:
-            if pod.status.phase == 'Running' and pod.status.host_ip in self._nodes_by_ip:
+            if (pod.status.phase == 'Running' or pod.status.phase == 'Pending') \
```
drmorr0: could we make this a set?  e.g.
```KUBERNETES_SCHEDULED_PHASES = {'Pending', 'Running'}```
and then we can just check `if pod.status.phase in KUBERNETES_SCHEDULED_PHASES`

**Review by drmorr0 — CHANGES_REQUESTED** [review]
This is a dumb minor nitpick but I think it's a little clearer this way.

File: clusterman/kubernetes/kubernetes_cluster_connector.py [hunk 2]
```diff
@@ -217,10 +217,12 @@ def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
         }
 
     def _get_pods_by_ip(self) -> Mapping[str, List[KubernetesPod]]:
+        KUBERNETES_SCHEDULED_PHASES = {'Pending', 'Running'}
```
drmorr0: fix'n'ship: any chance we could make this a global constant?

drmorr0 (approved):
fix'n'ship

PR #69, opened 2020-10-15
Change gpu resource name based on scheduler

mtbentley:
### Description

Depending on the scheduler, select either `gpus` or `nvidia.com/gpu` for gpu metrics

### Testing Done

none yet


mtbentley:
this is actually incorrect, this needs to be implemented in https://github.com/Yelp/clusterman/blob/master/clusterman/kubernetes/util.py#L56

PR #73, opened 2020-10-16
Restrict scaling down if we recently lost capacity for any reason

JoeMalt:
### Description
[CLUSTERMAN-553](https://jira.yelpcorp.com/browse/CLUSTERMAN-553) (Yelp-internal) has more context.

This adds an option to prevent scaling down if we've recently lost instances in the pool. Unlike existing limits on scaling down, this accoutns for instances going away for any reason, including k8s problems, Spot terminations, etc. 

It is an **extremely rough draft** and hasn't been tested yet, putting this PR out to make sure I'm going in the right direction.

### Testing Done
`make test` passes but I haven't actually tested this feature.

### DIFF
--- a/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
+++ b/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
@@ -16,7 +16,6 @@ autoscale_signal:
   internal: true
   period_minutes: 1
 
-# TODO (jmalt): this config option is a first draft, both naming and location are open to change
 autoscaling:
   prevent_scale_down_after_capacity_loss: true
   capacity_loss_threshold: 3
--- a/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
+++ b/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
@@ -15,3 +15,8 @@ scaling_limits:
 autoscale_signal:
   internal: true
   period_minutes: 1
+
+# TODO (jmalt): this config option is a first draft, both naming and location are open to change
+autoscaling:
+  prevent_scale_down_after_capacity_loss: true
+  capacity_loss_threshold: 3
--- a/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
+++ b/acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
@@ -18,4 +18,4 @@ autoscale_signal:
 
 autoscaling:
   prevent_scale_down_after_capacity_loss: true
-  capacity_loss_threshold: 3
+  instance_loss_threshold: 3
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -156,7 +156,7 @@ class Autoscaler:
         no_scale_down = False
         if self.autoscaling_config.prevent_scale_down_after_capacity_loss:
             removed_nodes_since_last_reload = self.pool_manager.get_removed_nodes_since_last_reload()
-            # TODO (jmalt): we may want something more sophisticated than len(nodes),
+            # TODO (jmalt, CLUSTERMAN-553): we may want something more sophisticated than len(nodes),
             # especially for heterogeneous SFRs
             if len(removed_nodes_since_last_reload) > self.autoscaling_config.capacity_loss_threshold:
                 logger.warning('Nodes lost since last autoscaler run is greater than limit,'
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -153,17 +153,17 @@ class Autoscaler:
 
         logger.info(f'Signal {signal_name} requested {resource_request}')
 
+        self.pool_manager.reload_state()
+
         no_scale_down = False
         if self.autoscaling_config.prevent_scale_down_after_capacity_loss:
-            removed_nodes_since_last_reload = self.pool_manager.get_removed_nodes_since_last_reload()
-            # TODO (jmalt, CLUSTERMAN-553): we may want something more sophisticated than len(nodes),
-            # especially for heterogeneous SFRs
-            if len(removed_nodes_since_last_reload) > self.autoscaling_config.capacity_loss_threshold:
+            removed_nodes_before_last_reload = self.pool_manager.get_removed_nodes_before_last_reload()
+            # TODO (CLUSTERMAN-576): support instance weights here, instead of just instance count
+            if len(removed_nodes_before_last_reload) > self.autoscaling_config.instance_loss_threshold:
                 logger.warning('Nodes lost since last autoscaler run is greater than limit,'
                                ' will not scale down on this run')
                 no_scale_down = True
 
-        self.pool_manager.reload_state()
         if isinstance(resource_request, list):
             pass
         else:
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -152,6 +152,17 @@ class Autoscaler:
             exception, tb = e, traceback.format_exc()
 
         logger.info(f'Signal {signal_name} requested {resource_request}')
+
+        no_scale_down = False
+        if self.autoscaling_config.prevent_scale_down_after_capacity_loss:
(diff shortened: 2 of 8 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -71,6 +72,23 @@ def reload_state(self) -> None:
         self._nodes_by_ip = self._get_nodes_by_ip()
         self._pods_by_ip = self._get_pods_by_ip()
 
+    def get_removed_nodes_since_last_reload(self) -> Set[KubernetesNode]:
+        try:
```
drmorr0: Can you rebase on master?  We now store a copy of the Kubernetes client object, so we shouldn't need to do this `try/except` block anymore.

clusterman/interfaces/cluster_connector.py
```diff
@@ -35,6 +37,10 @@ def reload_state(self) -> None:  # pragma: no cover
         """ Refresh any state that needs to be stored at the start of an autoscaling run """
         pass
 
+    def get_removed_nodes_since_last_reload(self) -> Set[KubernetesNode]:
```
drmorr0: I am on the fence about whether this should be included here if it's only supported for one subclass.  I get that it's here to pass mypy presumably, but maybe there's another way to do it that's cleaner?

Not too sure, if there's not a better way to do this, feel free to leave it alone.

acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
```diff
@@ -15,3 +15,8 @@ scaling_limits:
 autoscale_signal:
   internal: true
   period_minutes: 1
+
+# TODO (jmalt): this config option is a first draft, both naming and location are open to change
```
drmorr0: I'm happy with these names

drmorr0 (changes requested):
I think this is generally looking good; there's a couple minor nitpicks and a couple TODO items to clean up, but this seems OK to me!

clusterman/interfaces/cluster_connector.py
JoeMalt: It looks like mypy is happy without it, removed

JoeMalt:
@drmorr0 thanks for the review - if this is good to go, can you give me some pointers on testing it?

clusterman/__init__.py
```diff
@@ -11,4 +11,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-__version__ = '4.5.0'
```
drmorr0: I'm a bit confused here, this should already be on master, why is it showing up on your branch?  I think maybe something went funky with your rebase?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -49,28 +92,28 @@ class KubernetesClusterConnector(ClusterConnector):
...
-            raise
 
-        self._core_api = kubernetes.client.CoreV1Api()
+        self._core_api = CachedCoreV1Api(self.kubeconfig_path)
         self._pods = self._get_all_pods()
         self._nodes_by_ip = self._get_nodes_by_ip()
         self._pods_by_ip = self._get_pods_by_ip()
 
+    def get_removed_nodes_since_last_reload(self) -> Set[KubernetesNode]:
+        previous_nodes = self._nodes_by_ip
```
drmorr0: I think what I recommend here is storing the list of previous nodes above (before line 106) and then we can just return the set difference between previous nodes and current nodes, we don't have to call `_get_nodes_by_ip` twice.

acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
```diff
@@ -15,3 +15,7 @@ scaling_limits:
 autoscale_signal:
   internal: true
   period_minutes: 1
+
+autoscaling:
+  prevent_scale_down_after_capacity_loss: true
+  capacity_loss_threshold: 3
```
drmorr0: since this is _not_ weight-based (and almost everything else is) can we rename this to make it clear that it's based on the number of EC2 instances lost?

clusterman/autoscaler/autoscaler.py
```diff
@@ -152,6 +152,17 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             exception, tb = e, traceback.format_exc()
 
         logger.info(f'Signal {signal_name} requested {resource_request}')
+
+        no_scale_down = False
+        if self.autoscaling_config.prevent_scale_down_after_capacity_loss:
+            removed_nodes_since_last_reload = self.pool_manager.get_removed_nodes_since_last_reload()
+            # TODO (jmalt, CLUSTERMAN-553): we may want something more sophisticated than len(nodes),
```
drmorr0: Can we file a new ticket here that's specific to "making this use weights"?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -49,28 +92,28 @@ class KubernetesClusterConnector(ClusterConnector):
...
+        current_nodes = self._get_nodes_by_ip()
+
+        # todo jmalt: make sure this gives the desired results
```
drmorr0, line 78: I think this todo still needs addressing?  since we're literally just comparing the length above, you _could_ just calculate the length of the difference here, or possibly reduce this to a list of IPs?  I think it's also fine the way it is, up to you.

drmorr0 (changes requested):
I have a few more comments below.  As far as testing goes, testing on infrastage sounds good, but I also think we want to have an itest to verify the behaviour.

General docs on behave are here: https://behave.readthedocs.io/en/latest/

acceptance/srv-configs/clusterman-clusters/local-dev/default.kubernetes
JoeMalt: Will rename to `instance_loss_threshold`

clusterman/autoscaler/autoscaler.py
JoeMalt: Filed CLUSTERMAN-576

itests/steps/autoscaler.py
```diff
@@ -339,19 +372,25 @@ def autoscaler_runs(context):
...
+
 @behave.then('the autoscaler should scale rg(?P<rg>[12]) to (?P<target>\d+) capacity')
 def rg_capacity_change(context, rg, target):
     groups = list(context.autoscaler.pool_manager.resource_groups.values())
     if int(target) != groups[int(rg) - 1].target_capacity:
         assert_that(
             groups[int(rg) - 1].modify_target_capacity.call_args_list,
-            contains(
-                mock.call(int(target), dry_run=False),
+            has_item(
```
JoeMalt, line 392: FYI: I had to change this because sometimes we now only have one call to `modify_target_capacity`.

itests/steps/autoscaler.py
```diff
@@ -283,6 +295,15 @@ def k8s_autoscaler(context):
         monitoring_enabled=False,
     )
 
+    if prevent_scale_down_after_capacity_loss:
```
JoeMalt, line 297: I appreciate that this isn't as nice as overriding the config with`staticconf.test.PatchConfiguration` but I couldn't get that to override the settings defined in `itests/environment.py`. If this is too hacky, I'll have another go at getting PatchConfiguration to work.

clusterman/kubernetes/kubernetes_cluster_connector.py
JoeMalt: I think I'll keep it as-is, but this does raise a question of what happens if we lose and gain nodes in between autoscaler runs (e.g. some instances die and AWS replaces them). With this scenario, we'd consider that a capacity loss, but maybe we don't want to? If we don't, then maybe something like `len(prev_nodes) - len(current_nodes)` would be better.

What do you think?

JoeMalt: As for making sure that this functionality actually works end-to-end: the cluster connector is mocked in the itests so it's tricky to test there; I think I'll need to test it on dev somewhere (the infrastage Flink pool would do fine).

PR #74, opened 2020-10-16
cache calls to the Kubernetes API server for speed and profit

drmorr0:
### Description

Stick a caching layer in front of the apiserver to cache "expensive" calls that we don't want to repeat.

This was straightforward for the case of listing pods, and I minorly refactored the "listing nodes" behavior so we only have to make that call once, instead of once per pool.

### Testing Done

`make test` passes
manual testing on devc works, and seems to be returning the correct metrics values.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -22,6 +22,8 @@ from typing import Tuple
 import colorlog
 import kubernetes
 import staticconf
+from cachetools import TTLCache
+from cachetools.keys import hashkey
 from kubernetes.client.models.v1_node import V1Node as KubernetesNode
 from kubernetes.client.models.v1_node_selector_requirement import V1NodeSelectorRequirement
 from kubernetes.client.models.v1_node_selector_term import V1NodeSelectorTerm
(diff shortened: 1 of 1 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -38,6 +40,45 @@
...
+        except TypeError:
+            error_msg = 'Could not load KUBECONFIG; is this running on Kubernetes master?'
+            if 'yelpcorp' in socket.getfqdn():
+                error_msg += '\nHint: try using the clusterman-k8s-<clustername> wrapper script!'
+            logger.error(error_msg)
+            raise
+
+        self._client = kubernetes.client.CoreV1Api()
+
+    def __getattr__(self, attr):
```
EvanKrall, line 63: IMO it would be cleaner to not do this as `__getattr__` but instead to subclass from CoreV1Api and decorate overridden methods that simply call super():

```
class CachedCoreV1Api(CoreV1Api):
    ...
    @cachetools.func.ttl_cache
    def list_node(self, *args, **kwargs):  # or the actual arguments
        super().list_node(*args, **kwargs)

    @cachetools.func.ttl_cache
    def list_pod_for_all_namespaces(self, *args, **kwargs):  # or the actual arguments
        super().list_node(*args, **kwargs)
```

or

```
class CachedCoreV1Api(CoreV1Api):
    ...
    list_node =  cachetools.func.ttl_cache(CoreV1Api.list_node)
    list_pod_for_all_namespaces = cachetools.func.ttl_cache(CoreV1Api.list_pod_for_all_namespaces)
```


You could still override `__init__` to call `kubernetes.config.load_kube_config(kubeconfig_path)`.

EvanKrall: oh I guess these wouldn't work, as `self` would get passed in to the methods, so different instances wouldn't share the same cache.

EvanKrall (approved):

Rob-Johnson:
for Mesos we did this at the infra tier and ran nginx in front of the mesos api - maybe we can do the same thing for k8s rather than re-inventing a caching in every client?

drmorr0:
> for Mesos we did this at the infra tier and ran nginx in front of the mesos api - maybe we can do the same thing for k8s rather than re-inventing a caching in every client?

That's not a bad idea, though I wonder how certificates and things like RBAC will work?  Is there any possibility someone could read something from the cache that they're not supposed to have access to or something?  Anyways, I'll create a ticket for this.

PR #76, opened 2020-10-21
only cache calls to the API server if the right environment variable is set

drmorr0:
### Description

There's no reason to cache results from the api server in the autoscaler, because we're not querying it multiple times per autoscaling run.  So we'll only use the cache if we set the `KUBE_CACHE_ENABLED` environment variable.  I'm using an environment variable as opposed to adding something to srv-configs because there's not a _super_-easy way to make the config value only apply to the metrics collector.  Once this is shipped we'll need to modify the environment in yelpsoa for the metrics collectors.

### Testing Done

None.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 import socket
 from collections import defaultdict
 from distutils.util import strtobool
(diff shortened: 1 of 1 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -64,7 +65,7 @@ def __getattr__(self, attr):
         global KUBERNETES_API_CACHE
         func = getattr(self._client, attr)
 
-        if attr in self.CACHED_FUNCTION_CALLS:
+        if os.environ.get('KUBE_CACHE_ENABLED', '') and attr in self.CACHED_FUNCTION_CALLS:
```
stug, line 68: I think this is roughly ok, but it feels kind of gross to me to be doing this somewhere kind of deep in the code.  How hard would it be to follow the pattern of, e.g., the `CMAN_CLUSTER` environment variable and do this as an arg?

stug (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
drmorr0: Discussed on slack -- using args would require passing a bunch of junk through, which feels ugly in its own way.

PR #79, opened 2020-10-28
Pin cryptography version to 2.8 in acceptance tox env.  Later version…

stug:
…s are unhappy with the version of openssl we're using, which I I think would mean getting off a xenial base docker image.

### Description

The (internal) build is failing because cryptography 3.2 doesn't like openssl 1.0.2 (with reason -- it's no longer supported).  But we can't easily get newer versions until we're on a bionic base image.

### Testing Done

acceptance tests pass


drmorr0 (approved):

drmorr0 (commented):
lgtm

PR #81, opened 2020-11-02
make some changes to try to reduce memory usage

drmorr0:
### Description

The main change here is that we don't require the autoscalers to track
all the pods, just the ones that belong to the current pool.  This
should hopefully reduce the amount of memory used by _most_ of the
autoscaler instances.

To accomplish this, we just filter all the pods that are a) running on a
node in our pool, b) has a node selector for nodes in our pool, or c)
has an affinity for nodes in our pool.

We also only store the pods _once_, instead of having a list of all the
pods, and then have a separate dict of pods by IP.  (Actually I'm not sure, 
maybe we were just storing references in one of those, but anyways, 
*shrug*).

We still cache everything for the metrics collector, so these will still
need large memory utilization.

I *think* this maybe also solves [CLUSTERMAN-562](https://jira.yelpcorp.com/browse/CLUSTERMAN-562); @EvanKrall can you
take a look and see if you agree?

### Testing Done

Made sure all the tests still pass

### DIFF
--- a/tests/kubernetes/kubernetes_cluster_connector_test.py
+++ b/tests/kubernetes/kubernetes_cluster_connector_test.py
@@ -25,21 +25,71 @@ from kubernetes.client import V1PodStatus
 from kubernetes.client import V1ResourceRequirements
 from kubernetes.client.models.v1_affinity import V1Affinity
 from kubernetes.client.models.v1_node import V1Node as KubernetesNode
+from kubernetes.client.models.v1_node_address import V1NodeAddress
 from kubernetes.client.models.v1_node_affinity import V1NodeAffinity
 from kubernetes.client.models.v1_node_selector import V1NodeSelector
 from kubernetes.client.models.v1_node_selector_requirement import V1NodeSelectorRequirement
 from kubernetes.client.models.v1_node_selector_term import V1NodeSelectorTerm
 from kubernetes.client.models.v1_preferred_scheduling_term import V1PreferredSchedulingTerm
+from staticconf.testing import PatchConfiguration
 
 from clusterman.interfaces.types import AgentState
 from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
 
 
 @pytest.fixture
-def pod1():
+def running_pod_1():
     return V1Pod(
-        metadata=V1ObjectMeta(name='pod1'),
-        status=V1PodStatus(phase='Running'),
+        metadata=V1ObjectMeta(name='running_pod_1'),
+        status=V1PodStatus(phase='Running', host_ip='10.10.10.2'),
+        spec=V1PodSpec(containers=[
+               V1Container(
+                    name='container1',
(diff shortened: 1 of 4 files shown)


stug (approved):
I think this makes rough sense.  Will it be a problem if a pod that ends up running on a node in the pool is somehing that either runs anywhere on the cluster or runs on every node?  I kind of think not, but might be worth confirming?

drmorr0:
> I think this makes rough sense. Will it be a problem if a pod that ends up running on a node in the pool is somehing that either runs anywhere on the cluster or runs on every node? I kind of think not, but might be worth confirming?

So, if the pod is already running on the pool, it will get accounted for correctly (it will appear as using resources on the pool).  If the pod is `Pending` or in some other state where it has not been scheduled on the pool, it will only get accounted for if it has a `nodeSelector` which matches the poolname, or if it has a `nodeAffinity` which matches the poolname.  

Things that can run literally anywhere won't get accounted for by this pool's (or any) autoscaler instance, but I believe that is true in the status quo as well.  Things that run everywhere (e.g., `kiam` or other DaemonSets) have nodeSelectors that match the node name, and will also not get accounted for by any autoscaler instance, but that is also the status quo.  An easy improvement here is to add a nodeSelector check for the node name.

EvanKrall (approved):

tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -278,38 +345,28 @@ def test_delete_node(mock_cluster_connector):
...
+        ['not-bar']
+    assert mock_cluster_connector._pod_belongs_to_pool(running_pod_1)
+    assert mock_cluster_connector._pod_belongs_to_pool(running_pod_2)
+    assert mock_cluster_connector._pod_belongs_to_pool(unevictable_pod)
+    assert mock_cluster_connector._pod_belongs_to_pool(unschedulable_pod)
+    assert mock_cluster_connector._pod_belongs_to_pool(pending_pod)
+    assert mock_cluster_connector._pod_belongs_to_pool(pod_with_required_affinity)
+    assert mock_cluster_connector._pod_belongs_to_pool(pod_with_preferred_affinity)
+    assert not mock_cluster_connector._pod_belongs_to_pool(pod_with_node_selector_elsewhere)
+    assert not mock_cluster_connector._pod_belongs_to_pool(pod_with_required_affinity_elsewhere)
```
EvanKrall, line 374: can you add an assertion here for `running_pod_on_nonexistent_node`?

Otherwise I'm not seeing a test that actually asserts that these pods are counted.

drmorr0: They are counted in the test for getting allocated resources, but I will also add a check here.

stug (approved):
lg2m other than a comment change in `_pod_belongs_to_pool` to clarify what'll happen for CLUSTERMAN-562

PR #82, opened 2020-11-04
Drmorr/get test coverage back up

drmorr0:
### Description

This is basically a no-op change, I just added some more tests to get our coverage back up (and I think they're more-or-less useful tests, as well).

The two substantive changes I made were to 
* delete a bunch of dead code around evicting pods and cordoning nodes from the `KubernetesClusterConnector`; we don't do this from inside Clusterman anymore, instead it should be handled from `graceful_k8s_shutdown`.
* I don't think we should just return `0` if kubernetes returns a bad value for gpu count, I think it's better to just crash in this case.

### Testing Done

`make test` passes

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -206,11 +206,12 @@ class PoolManager:
             for instance_metadata in group.get_instance_metadatas(state_filter)
         ]
 
+    # currently dead code, so don't count towards coverage metrics
     def _filter_scale_up_options_for_pod(
         self,
         pod: KubernetesPod,
         scale_up_options: Mapping[str, List[ClusterNodeMetadata]],
-    ) -> Mapping[str, List[ClusterNodeMetadata]]:
+    ) -> Mapping[str, List[ClusterNodeMetadata]]:  # pragma: no cover
         filtered_options: Mapping[str, List[ClusterNodeMetadata]] = defaultdict(list)
         for group_id, options in scale_up_options.items():
             for option in options:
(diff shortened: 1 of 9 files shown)


clusterman/autoscaler/pool_manager.py
```diff
@@ -206,11 +206,12 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+    # currently dead code, so don't count towards coverage metrics
     def _filter_scale_up_options_for_pod(
         self,
         pod: KubernetesPod,
         scale_up_options: Mapping[str, List[ClusterNodeMetadata]],
-    ) -> Mapping[str, List[ClusterNodeMetadata]]:
+    ) -> Mapping[str, List[ClusterNodeMetadata]]:  # pragma: no cover
         filtered_options: Mapping[str, List[ClusterNodeMetadata]] = defaultdict(list)
         for group_id, options in scale_up_options.items():
             for option in options:
```
stug, line 217: why not just delete it?

stug (approved):
lg2m if we're sure its dead

clusterman/autoscaler/pool_manager.py
drmorr0: In the _super ideal_ world this code would become un-dead at some point in the future.

drmorr0:
> lg2m if we're sure its dead

```
(virtualenv_run) drmorr@dev54-uswest1adevc ~cman X git --no-pager grep set_node_unschedulable
(virtualenv_run) drmorr@dev54-uswest1adevc ~cman X git --no-pager grep evict_pods_on_node
(virtualenv_run) drmorr@dev54-uswest1adevc ~cman X git --no-pager grep delete_node
```

PR #83, opened 2020-11-06
Add debug logging to autoscaler.py

JoeMalt:
### Description

The `prevent_scale_down_after_capacity_loss` option didn't seem to work when I tested it in infrastage so I need to have another go, with some logging added. I'll revert this change once I'm done.

### Testing Done

`make test` passes. This only adds log lines anyway.


AshishKhatkar (approved):
lgtm

PR #86, opened 2020-11-09
fix servicedocs building and uploading

JoeMalt:
### Description

I had to make these changes in order to upload changes to servicedocs. I've probably missed something that's already configured (after all, past changes must have been uploaded somehow), so feel free to discard this if it's not actually needed.

In no particular order:

- change `Sphinx` to a version that exists in Yelp's internal PyPi
- add an `upload_docs` tox target
- install `servicedocs-tool` which provides the `upload-to-servicedocs` utility

### Testing Done

none, but uploading docs works now

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -6,8 +6,9 @@ MarkupSafe==1.1.1
 packaging==19.2
 Pygments==2.4.2
 pytz==2019.3
+servicedocs-tool==2.4.4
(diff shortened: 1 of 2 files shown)


JoeMalt:
Canonical docs are on https://clusterman.readthedocs.io/en/latest/configuration.html which updates automatically from GitHub, so this isn't needed.

PR #88, opened 2020-11-10
Drmorr/fix some clusterman bugs

drmorr0:
### Description

This change fixes two bugs found in the autoscaling behaviour

1) The first, and less serious, is that we weren't reloading the pool
state until _after_ we'd already evaluated our signals, which means that
for the internal version of the signal (which doesn't rely on the
metrics store), we were using the data from the _last_ autoscaling run
instead of the current one.

2) The second is a bit larger: it turns out we were never actually
scaling based on pending pods, because pending pods were always getting
marked as "Unknown".  The reason for this is that we were comparing the
pending pods request to the total allocatable capacity on the node,
instead of what's remaining, so Clusterman never realized that pods
potentially couldn't fit on a given node.

This change fixes both of these behaviours.  Since these both change how
the scaling behaviour actually happens, we'll want to roll it out
slowly.

### Testing Done

I fixed the itests that check scaling behaviour to actually make sure the 
pods are getting classified correctly.

### DIFF
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -141,6 +141,7 @@ class Autoscaler:
             logger.info('Autoscaling is currently paused; doing nothing')
             return
 
+        self.pool_manager.reload_state()
         try:
             signal_name = self.signal.name
             resource_request = self.signal.evaluate(timestamp)
@@ -152,8 +153,6 @@ class Autoscaler:
             exception, tb = e, traceback.format_exc()
 
         logger.info(f'Signal {signal_name} requested {resource_request}')
-
-        self.pool_manager.reload_state()
 
         no_scale_down = False
         if self.autoscaling_config.prevent_scale_down_after_capacity_loss:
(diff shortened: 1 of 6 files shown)


clusterman/autoscaler/autoscaler.py
```diff
@@ -141,6 +141,7 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             logger.info('Autoscaling is currently paused; doing nothing')
             return
 
+        self.pool_manager.reload_state()
```
kawaiwanyelp, line 144: nit: I'd add a comment here about this needing to happen before the pool signal is evaluated, especially we didn't catch this before.

kawaiwanyelp (approved):
Fixes look good. I assume you'll roll this out to stage first, then one prod at a time.

clusterman/autoscaler/autoscaler.py
drmorr0: I did one better and added a test to catch this.

kawaiwanyelp (approved):

PR #89, opened 2020-11-25
add some logging for the pending pods signal

drmorr0:
### Description

It's a little hard to see sometimes what Clusterman thinks is pending and why, and what it's doing with those resources, so this adds a couple log lines to make this more clear.

### Testing Done

none

### DIFF
--- a/clusterman/signals/pending_pods_signal.py
+++ b/clusterman/signals/pending_pods_signal.py
@@ -35,6 +35,9 @@ def _get_resource_request(
                 # this with a more intelligent solution in the future.
                 resource_request += total_pod_resources(pod) * 2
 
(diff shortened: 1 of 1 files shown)


clusterman/signals/pending_pods_signal.py
```diff
@@ -35,6 +35,9 @@ def _get_resource_request(
                 # this with a more intelligent solution in the future.
                 resource_request += total_pod_resources(pod) * 2
 
+    logger.info(f'List of pending pods: {pending_pods}')
```
stug, line 38: Will this include the pod resource requests/limits?

stug (approved):
lg2m, one question/thought about what we're logging

PR #93, opened 2020-12-30
CLUSTERMAN-542: Allow excluding daemonset pods from resource calculations

jfongatyelp:
### Description

We saw in CLUSTERMAN-542 that clusterman refused to remove hosts running nothing but pods common to every node running from daemonsets because these pods were counted towards *allocated* resources requested.

This change adds an optional pool-level toggle `exclude_daemonset_pods` that will exclude any pods created via daemonsets from being added to *allocated* or *requested* resources, and any resources they are taking will be removed from *total* available resources.  This will take both `Pending` and `Running` daemonset pods into account on the appropriate pool nodes.  

This will also exclude daemonset pods from being counted towards running tasks taken into consideration for killing nodes (i.e. daemonset pods aren't counted towards `max_tasks_to_kill`), as task count is taken directly from `_pods_by_ip`, which would now exclude all daemonset pods with this config toggle.

### Testing Done
Added a test into `kubernetes_cluster_connector_test` to verify this config and calculations do what we expect, and ran manually against `kubestage`'s `default` pool:

**Without `exclude_daemonset_pods`**:
```
INFO:clusterman.autoscaler.autoscaler:Current cluster total resources: ClustermanResources(cpus=156.0, mem=297769.544704, disk=2242208.257206, gpus=0)
INFO:clusterman.autoscaler.autoscaler:Current cluster allocated resources: ClustermanResources(cpus=8.699999999999998, mem=46628.0, disk=19750.0, gpus=0)
```

**With `exclude_daemonset_pods`**:
```
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0.7000000000000001 cpus from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 1300.0 mem from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 100.0 disk from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0 gpus from daemonset pods
INFO:clusterman.autoscaler.autoscaler:Current cluster total resources: ClustermanResources(cpus=155.3, mem=296469.544704, disk=2242108.257206, gpus=0)
INFO:clusterman.autoscaler.autoscaler:Current cluster allocated resources: ClustermanResources(cpus=7.999999999999998, mem=45328.0, disk=19650.0, gpus=0)
```

After increasing target capacity in kubestage, then manually modifying local `default.kubernetes` to set `max_tasks_to_kill: 1`, a dry-run with `exclude_daemonset_pods: true` would have scaled down appropriately:
```
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0.6000000000000001 cpus from total resources
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 1000.0 mem from total resources
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 200.0 disk from total resources
...
INFO:clusterman.autoscaler.autoscaler:Current cluster total resources: ClustermanResources(cpus=87.4, mem=176373.6448, disk=1824558.235394, gpus=0)
INFO:clusterman.autoscaler.autoscaler:Current cluster allocated resources: ClustermanResources(cpus=12.2, mem=70328.0, disk=24404.0, gpus=0)
...
INFO:clusterman.autoscaler.pool_manager:Killable instance IDs in kill order:
['i-09cec463628b2136b', 'i-03b9e0f50e8956c7f']
```

without `exclude_daemonset_pods`, the autoscaler found 0 killable instances:
```
INFO:clusterman.autoscaler.autoscaler:Current cluster total resources: ClustermanResources(cpus=88.0, mem=177373.6448, disk=1824758.235394, gpus=0)
INFO:clusterman.autoscaler.autoscaler:Current cluster allocated resources: ClustermanResources(cpus=12.799999999999997, mem=71328.0, disk=24604.0, gpus=0)
...
INFO:clusterman.autoscaler.pool_manager:Killable instance IDs in kill order:
[]
```

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -97,7 +97,7 @@ class KubernetesClusterConnector(ClusterConnector):
             for node in self._nodes_by_ip.values()
         )
         if excluded_total > 0:
-            logger.debug(f'Excluded {excluded_total} {resource_name} from total resources by config')
+            logger.info(f'Excluded {excluded_total} {resource_name} from total resources')
         return base_total - excluded_total
 
     def get_unschedulable_pods(self) -> List[Tuple[KubernetesPod, PodUnschedulableReason]]:
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -87,18 +87,18 @@ class KubernetesClusterConnector(ClusterConnector):
         )
 
     def get_resource_total(self, resource_name: str) -> float:
-        # Total node resources minus any allocated resources to excluded pods
-        base_total = sum(
-            getattr(total_node_resources(node), resource_name)
-            for node in self._nodes_by_ip.values()
+        if self._excluded_pods_by_ip:
+            logger.info(f'Excluded {self.get_resource_excluded(resource_name)} {resource_name} from daemonset pods')
+        return sum(
+            getattr(total_node_resources(node, self._excluded_pods_by_ip.get(node_ip, [])), resource_name)
+            for node_ip, node in self._nodes_by_ip.items()
         )
-        excluded_total = sum(
-            getattr(allocated_node_resources(self._excluded_pods_by_ip[get_node_ip(node)]), resource_name)
-            for node in self._nodes_by_ip.values()
+
+    def get_resource_excluded(self, resource_name: str) -> float:
+        return sum(
+            getattr(allocated_node_resources(self._excluded_pods_by_ip.get(node_ip, [])), resource_name)
+            for node_ip in self._nodes_by_ip.keys()
         )
-        if excluded_total > 0:
-            logger.info(f'Excluded {excluded_total} {resource_name} from total resources')
-        return base_total - excluded_total
 
     def get_unschedulable_pods(self) -> List[Tuple[KubernetesPod, PodUnschedulableReason]]:
         unschedulable_pods = []
@@ -155,8 +155,11 @@ class KubernetesClusterConnector(ClusterConnector):
         pod_resource_request = total_pod_resources(pod)
         for node_ip, pods_on_node in self._pods_by_ip.items():
             node = self._nodes_by_ip.get(node_ip)
-            if node and pod_resource_request < total_node_resources(node) - allocated_node_resources(pods_on_node):
-                return PodUnschedulableReason.Unknown
+            if node:
+                available_node_resources = (total_node_resources(node, self._excluded_pods_by_ip.get(
+                    node_ip, [])) - allocated_node_resources(pods_on_node))
+                if pod_resource_request < available_node_resources:
+                    return PodUnschedulableReason.Unknown
 
         return PodUnschedulableReason.InsufficientResources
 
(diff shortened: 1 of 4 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,10 +87,18 @@ def get_resource_allocation(self, resource_name: str) -> float:
         )
 
     def get_resource_total(self, resource_name: str) -> float:
-        return sum(
+        # Total node resources minus any allocated resources to excluded pods
```
thebostik: This makes it clear what this code does, is there a reason we want task count and resources to stay in sync like this  with the handling of the excluded resources? If so, I'd imagine this potentially working with the code/config that defines total node resources, like would we benefit from a one-time decrease by 0.6cpu/1000mb (static) or `sum(excluded_resources)` (dynamic) to offset this?

thebostik (approved):
this lgtm! one question in here about how i think this will lead to an increase in per-host capacity, which might be fine, or might be worth correcting.

Also, please remember to revert https://github.yelpcorp.com/sysgit/srv-configs/pull/2272 before shipping this or else hosts in those pools might be killed with the new "higher" limit!

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -86,10 +87,18 @@ def get_resource_allocation(self, resource_name: str) -> float:
...
+        base_total = sum(
             getattr(total_node_resources(node), resource_name)
             for node in self._nodes_by_ip.values()
         )
+        excluded_total = sum(
+            getattr(allocated_node_resources(self._excluded_pods_by_ip[get_node_ip(node)]), resource_name)
+            for node in self._nodes_by_ip.values()
+        )
+        if excluded_total > 0:
+            logger.info(f'Excluded {excluded_total} {resource_name} from total resources')
```
kawaiwanyelp: nit: I'd extend the info to say that the resources were excluded because they are being consumed by daemon sets to make it clear for debugging.

clusterman/kubernetes/kubernetes_cluster_connector.py
jfongatyelp: Ah you are absolutely right, I missed that we use the total node resources calculation elsewhere.  I agree it makes more sense to move this logic to that code instead.

kawaiwanyelp (approved):
This seems like a good change that should be applied to all clusters. Will you be doing a srv-configs controlled staged rollout?

qui (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
jfongatyelp: Updated!

Output now shows:
```
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0.7000000000000001 cpus from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 1300.0 mem from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 100.0 disk from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0 gpus from daemonset pods
```

clusterman/kubernetes/kubernetes_cluster_connector.py
jfongatyelp: The exclusion was moved to where we calculate total per-node resources, so all code that fetched "total" node resources now report only the total minus excluded resources consistently.

jfongatyelp:
> This seems like a good change that should be applied to all clusters. Will you be doing a srv-configs controlled staged rollout?

Yes, exactly. Since this pulls from a new config toggle `exclude_daemonset_pods`, I plan to roll out in stages to make sure nothing unexpected happens during autoscaling, then eventually enable as default.

qui (approved):

thebostik (approved):
lgtm!

PR #96, opened 2021-01-12
CLUSTERMAN-542: Also allow us to specify default exclude_daemonset_po…

jfongatyelp:
…ds from clusterman config

### Description

In #93 I added the `exclude_daemonset_pods`, but did not read the default value from the primary config.  This fixes that, so we can toggle this at a wider level than just "everything" or "every individual pool".

### Testing Done

`make test` is green and a manual test with `exclude_daemonset_pods` excluded on the per-pool configuration but in `clusterman.yaml` excludes the daemonset resources as expected:
```
$ grep exclude_daemonset_pods /nail/srv/configs/cluster
exclude_daemonset_pods: true
nobody@e6be530dfc24:/code$ grep exclude_daemonset_pods /nail/srv/configs/clusterman-clusters/kubestage/default.kubernetes
nobody@e6be530dfc24:/code$
```

```
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0.4 cpus from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 600.0 mem from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 200.0 disk from daemonset pods
INFO:clusterman.kubernetes.kubernetes_cluster_connector:Excluded 0 gpus from daemonset pods
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,6 +152,8 @@ debug:
 		-v $(shell pwd)/.cman_debug_bashrc:/home/nobody/.bashrc:ro \
 		-v $(shell pwd)/etc-kubernetes:/etc/kubernetes:ro \
 		-v /nail/srv/configs:/nail/srv/configs:ro \
+		-v $(shell pwd)/clusterman.yaml:/nail/srv/configs/clusterman.yaml:ro \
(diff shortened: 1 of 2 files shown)


thebostik (approved):
lgtm!

qui (approved):

PR #101, opened 2021-01-18
Added new error exception when no resource groups are found

lanceleih:
### Description

Created a new Error Exception called `NoResourceGroupsFoundError`. This is created as per one of the action items of DAR-1046 to make Error Exception `AllResourceGroupsAreStaleError` more specific.

Clusterman will now first check if there are any resource groups at all. 

- If there are none, it will raise `NoResourceGroupsFoundError`. 
- If there are resource groups (but are considered stale), it will raise `AllResourceGroupsAreStaleError`.

### Testing Done

`make test - COMPLETE`

### DIFF
--- a/tests/autoscaler/pool_manager_test.py
+++ b/tests/autoscaler/pool_manager_test.py
@@ -505,7 +505,7 @@ def test_compute_new_resource_group_targets_scale_up_stale_pools_0(non_stale_cap
 
 
 def test_compute_target_capacity_no_resource_groups_found(mock_pool_manager):
-    mock_pool_manager.resource_groups = []
+    mock_pool_manager.resource_groups = {}
     with pytest.raises(NoResourceGroupsFoundError):
         mock_pool_manager.target_capacity == 0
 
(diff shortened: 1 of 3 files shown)


tests/autoscaler/pool_manager_test.py
```diff
@@ -503,6 +504,12 @@ def test_compute_new_resource_group_targets_scale_up_stale_pools_0(non_stale_cap
     assert new_targets == {'sfr-0': 2, 'sfr-1': 2, 'sfr-2': 2, 'sfr-3': 0, 'sfr-4': 0, 'sfr-5': 0, 'sfr-6': 0}
 
 
+def test_compute_target_capacity_no_resource_groups_found(mock_pool_manager):
+    mock_pool_manager.resource_groups = []
```
stug: Although it doesn't _really_ matter for this test (since all that matters is that `resource_groups` is "Falsy"), this should be a dict so that it has the same type as it would in real operation.

stug (approved):
Looks good other than a small quibble I have about types -- please fix'n'ship

tests/autoscaler/pool_manager_test.py
lanceleih: Latest commit changes type to dict. Resolving.

PR #105, opened 2021-02-17
Check account ID when enabling or disabling clusters

TheNavigat:


clusterman/cli/toggle.py
```diff
@@ -22,20 +23,37 @@
...
 from clusterman.util import autoscaling_is_paused
 from clusterman.util import CLUSTERMAN_STATE_TABLE
 from clusterman.util import parse_time_string
 
+logger = colorlog.getLogger(__name__)
+
+
+@timeout_wrapper
+def check_account_id(cluster) -> None:
+    current_account_id = sts.get_caller_identity()
```
mks-m: i think response type for this endpoint isn't actually account number string, see https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

try running the command with your changes and verify if it works

clusterman/cli/toggle.py
```diff
@@ -22,20 +23,37 @@
...
+    cluster_account_id = staticconf.read_string(f'clusters.{cluster}.aws_account_number')
+
+    if(current_account_id != cluster_account_id):
+        logger.warning(
```
mks-m: i would raise an error instead of logging here, because disabling or enabling the cluster using wrong iam user is incorrect operation so we shouldn't even attempt doing it

if we do decide to raise an error, let's also rename the function from `check_...` to `ensure_...`

mks-m (changes requested):

clusterman/cli/toggle.py
TheNavigat: You're correct, I fixed it. Thanks :)

mks-m (approved):
lgtm

PR #108, opened 2021-04-08
Reduce time needed for clusterman status by using server side filtering

Buffer0x7cd:
### Description

Earlier, During invocation of `clusterman status`, `autoscaling_resource_group` will get the list of all available auto scaling groups in account and then loop through them for tag based filtering. In large accounts where number of ASGs are high, This can take longer time and eventually slow down the status command. 

This PR introduces changes  in `_get_resource_group_tags` method, where instead of getting list of all ASGs in an account, It uses `describe-tags` api to filter out only required  tags. Due to this server side filtering, `poolmanger` class will only need to go through limited number of ASGs even if the total number of ASGs are high in number.
working correctly.

### DIFF
--- a/clusterman/aws/auto_scaling_resource_group.py
+++ b/clusterman/aws/auto_scaling_resource_group.py
@@ -296,9 +296,27 @@ class AutoScalingResourceGroup(AWSResourceGroup):
 
     @classmethod
     @ttl_cache(ttl=RESOURCE_GROUP_CACHE_SECONDS)
-    def _get_resource_group_tags(cls) -> Mapping[str, Mapping[str, str]]:
+    def _get_resource_group_tags(cls, filter_tag: str = '') -> Mapping[str, Mapping[str, str]]:
         """ Retrieves the tags for each ASG """
         asg_id_to_tags = {}
+
+        if filter_tag:
+            filter = [
+                {
+                    'Name': 'key',
+                    'Values': [
+                        filter_tag,
+                    ]
+                },
+            ]
+            """If the filter tag is present switch to desribe-tag method for server side
+                filtering for performance reasons
+            """
+            for page in autoscaling.get_paginator('describe_tags').paginate(Filters=filter):
+                for instance in page['Tags']:
+                    asg_id_to_tags[instance['ResourceId']] = {filter_tag: instance['Value']}
+            return asg_id_to_tags
+
         for page in autoscaling.get_paginator('describe_auto_scaling_groups').paginate():
             for asg in page['AutoScalingGroups']:
                 tags_dict = {tag['Key']: tag['Value'] for tag in asg['Tags']}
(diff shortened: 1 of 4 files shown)


clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -296,9 +296,27 @@ def _target_capacity(self) -> float:
...
+                    'Name': 'key',
+                    'Values': [
+                        filter_tag,
+                    ]
+                },
+            ]
+            """If the filter tag is present switch to desribe-tag method for server side
+                filtering for performance reasons
+            """
+            for page in autoscaling.get_paginator('describe_tags').paginate(Filters=filter):
```
vkhromov, line 315: I assume that this iterate through all ASGs with the tag(s) described in `filter_tag`, right?

clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -296,9 +296,27 @@ def _target_capacity(self) -> float:
...
+                for instance in page['Tags']:
+                    asg_id_to_tags[instance['ResourceId']] = {filter_tag: instance['Value']}
```
vkhromov, line 317: Is it ok that here we just put `instance['Value']` into the dict, but below we're building a new dict?
Is `instance['Value']` is already in the desired format?

vkhromov (commented):
Looks good overall, just please see a couple of questions below.
Also, could you describe how have you tested the change to make sure that there is no regression?

clusterman/aws/auto_scaling_resource_group.py
Buffer0x7cd: Yeah,  Here we are iterating through all of the ASGs with filter tag, and building a new dict `{asg-id: filtertag}` format

clusterman/aws/auto_scaling_resource_group.py
Buffer0x7cd: The output expected from_get_resource_group_tags function is in below format
```
{
  asg-id: {tag-name: tag-value}
}
```
But the output from describe-tag is in below format 

```
        {
            "ResourceId": "asg-id",
            "ResourceType": "auto-scaling-group",
            "Key": "filterTagName",
            "Value": "Filter-tag value",
            "PropagateAtLaunch": true
        }
```

Due to this, On line 317, I am getting `Instance['Value']` and then creating the required dict in 
`{filter-tag: Instance["Value"]} `  format

clusterman/aws/auto_scaling_resource_group.py
Buffer0x7cd: For testing, I ran the itest suite after the changes, Also these changes keeps the return format from `_get_resource_group_tags` function similar

vkhromov (approved):

PR #110, opened 2021-04-13
This commit add describe_tags method in moto, which was causing earli…

Buffer0x7cd:
After making changes regarding `describe_tags` method, we noticed  that tests have started failing, due to underlying testing library `moto`, does not support `describe_tags` method on `autoscaling` client. This change adds a workaround, So that the change can go ahead without breaking all of the tests. 

The original  issue can be tracked here

https://github.com/spulec/moto/issues/3854


vkhromov (approved):

vkhromov:
Please make sure that all tests pass with this PR and your previous change.

PR #112, opened 2021-04-15
Remove explicit references to internal pypi server

TheNavigat:
`make test` and `make itest` both run successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,6 @@ RUN     apt-get update \
             zk-flock \
         && apt-get clean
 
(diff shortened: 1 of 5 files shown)


analogue:
Why does this project have both internal and external git repos?

TheNavigat:
> Why does this project have both internal and external git repos?

Good question.

I don't know. But the internal one is a mirror of this, AFAIK. @tzhu-yelp: do you have some context?

tzhu-yelp:
Because of paasta tags on the internal repo for service deployment. Regardless, this change is covered by #134

PR #113, opened 2021-04-16
Don't die when termination fails

TheNavigat:
Host could possibly be terminated for whatever reason. Just catch the exception and go on.

`make test` passes successfully.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -395,7 +395,11 @@ def terminate_host(host: Host) -> None:
     logger.info(f'Terminating: {host.instance_id}')
     resource_group_class = RESOURCE_GROUPS[host.sender]
     resource_group = resource_group_class(host.group_id)
-    resource_group.terminate_instances_by_id([host.instance_id])
+
+    try:
+        resource_group.terminate_instances_by_id([host.instance_id])
+    except Exception:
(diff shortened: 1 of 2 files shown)


clusterman/draining/queue.py
```diff
@@ -395,7 +395,11 @@ def terminate_host(host: Host) -> None:
     logger.info(f'Terminating: {host.instance_id}')
     resource_group_class = RESOURCE_GROUPS[host.sender]
     resource_group = resource_group_class(host.group_id)
-    resource_group.terminate_instances_by_id([host.instance_id])
+
+    try:
+        resource_group.terminate_instances_by_id([host.instance_id])
+    except Exception:
+        logger.error(f'Failed to terminate host: {host.hostname}, possibly already terminated.')
```
danielpops, line 402: nit: include the actual error text instead of printing an assumption of what went wrong?

danielpops, line 402: nit: is this just for human user interactions? in case of use by non-humans, where do these logs go? are they monitored?

TheNavigat: As far as I can tell these logs aren't processed automatically, and they stay on local host. The script is executed by machines but the logs exist for investigation and debugging if unexpected behaviour occurs. @kawaiwanyelp should have more context on this

PR #114, opened 2021-04-19
CEP-1416: black as formatter

nemacysts:
this was mostly all done by the internal formatting script: all I did was re-format the .pre-commit-config.yaml file + tox.ini + add some `noqa`

`make test itest` worked just fine


vkhromov (changes requested):
Please add `pyproject.toml` into the first commit.

nemacysts:
> Please add `pyproject.toml` into the first commit.

ah, good catch: i went ahead and made the max line length match the previous config too: https://github.com/Yelp/clusterman/pull/114/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 120
+target_version = ['py36']
```
vkhromov:
```
target_version = ['py37']
```
?

vkhromov (approved):
lgtm overall, just 1 nit

pyproject.toml
nemacysts: whoops!

PR #122, opened 2021-08-27
Add ability to get EC2 node tags and obtain the cordon tag

TheNavigat:
### Testing Done

`make test`

### DIFF
--- a/clusterman/aws/aws_resource_group.py
+++ b/clusterman/aws/aws_resource_group.py
@@ -29,6 +29,7 @@ import simplejson as json
 
 from clusterman.aws.client import ec2
 from clusterman.aws.client import ec2_describe_instances
+from clusterman.aws.client import InstanceDict
 from clusterman.aws.markets import get_instance_market
 from clusterman.aws.markets import InstanceMarket
 from clusterman.interfaces.resource_group import InstanceMetadata
@@ -79,12 +80,14 @@ class AWSResourceGroup(ResourceGroup, metaclass=ABCMeta):
             instance_market = get_instance_market(instance_dict)
             instance_ip = instance_dict.get("PrivateIpAddress")
             hostname = gethostbyaddr(instance_ip)[0] if instance_ip else None
+            is_cordoned = self._is_instance_cordoned(instance_dict)
 
             metadata = InstanceMetadata(
                 group_id=self.id,
                 hostname=hostname,
                 instance_id=instance_dict["InstanceId"],
                 ip_address=instance_ip,
+                is_cordoned=is_cordoned,
                 is_stale=(instance_dict["InstanceId"] in self.stale_instance_ids),
                 market=instance_market,
                 state=aws_state,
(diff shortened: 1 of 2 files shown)


clusterman/autoscaler/pool_manager.py
```diff
@@ -47,9 +47,9 @@
 from clusterman.monitoring_lib import get_monitoring_client
 from clusterman.util import read_int_or_inf
 
-AWS_RUNNING_STATES = ("running",)
+AWS_RUNNING_STATES = ('running',)
```
nemacysts: it looks like you might have an old checkout that doesn't include the `black` reformatting changes we did (or are somehow running black + old pre-commit hooks)

mks-m (changes requested):
can you split all the code styling changes from functionality changes into separate commits?

clusterman/aws/aws_resource_group.py
```diff
@@ -72,23 +72,37 @@ def __init__(self, group_id: str, **kwargs: Any) -> None:
...
+            # If EC2 node has no tags, None is returned.
+            # This sets the value to [] instead.
+            tags = instance_dict.get('Tags', [])
+
+            for tag in tags:
+                if tag['Key'] == 'cordon':
+                    if tag['Value'] == 'true':
+                        is_cordoned = True
+
+                    break
```
nemacysts: could this be a helper method?

```Python
def is_instance_cordoned(instance_dict: Mapping[str, str]) -> bool:
    tags = instance_dict.get('Tags', [])
    for tag in tags:
        if tag['Key'] == 'cordon' and tag['Value'] == 'true':
            return True
    return False
```

if we wanted to get really fancy, we could probably do something like `return any(tag['Key'] == 'cordon' and tag['Value'] == 'true' for tag in tags)` but that might be too code-golfy :p

clusterman/aws/aws_resource_group.py
```diff
@@ -94,6 +97,18 @@ def get_instance_metadatas(self, state_filter: Optional[Collection[str]] = None)
...
+        # This sets the value to [] instead.
+        tags = instance_dict.get("Tags", [])
+
+        for tag in tags:
+            if tag["Key"] == "cordon":
+                if tag["Value"] == "true":
+                    return True
+
+        return False
+
```
nemacysts, line 111: technically this doesn't need to be inside this class since it doesn't use any instance attributes - but that's pretty nitpicky so feel free to ignore

nemacysts (approved):

clusterman/aws/aws_resource_group.py
TheNavigat: That's a good call. I added the function to the class because it's relevant to it (note the leading `_` for example), although it never really uses `self` indeed. It's more of a design and relevancy choice rather than a technical requirement one. I'd rather keep things ordered that way, so I'm going to merge.

PR #123, opened 2021-09-02
Don't kill EC2 instances with is_cordoned: True tag

TheNavigat:
`make test`

This hasn't been actually tested with a node having this tag yet. I'm aiming to merge this first before doing that.

### DIFF
--- a/tests/autoscaler/pool_manager_test.py
+++ b/tests/autoscaler/pool_manager_test.py
@@ -37,6 +37,7 @@ def _make_metadata(
     tasks=5,
     batch_tasks=0,
     is_safe_to_kill=True,
+    is_cordoned=False,
 ):
     return ClusterNodeMetadata(
         AgentMetadata(
@@ -51,6 +52,7 @@ def _make_metadata(
             hostname="host1",
             instance_id=instance_id,
             ip_address="1.2.3.4",
+            is_cordoned=is_cordoned,
             is_stale=is_stale,
             market="market-1",
             state="running",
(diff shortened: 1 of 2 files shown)


tests/autoscaler/pool_manager_test.py
```diff
@@ -338,6 +340,15 @@ def test_killable_nodes_under_target_capacity(self, mock_logger, mock_pool_manag
...
         assert "would take us under our target_capacity" in mock_logger.info.call_args[0][0]
 
+    def test_killable_node_cordoned(self, mock_logger, mock_pool_manager):
+        mock_pool_manager.get_node_metadatas = mock.Mock(
+            return_value=[_make_metadata("sfr-1", "i-1", is_cordoned=True)]
+        )
+
+        mock_pool_manager.max_tasks_to_kill = 100
+
+        assert mock_pool_manager._choose_nodes_to_prune(100, None) == {}
```
Buffer0x7cd, line 350: It seems like here you are testing the return from an internal function , If possible we should avoid it. maybe you can try mocking the draining_client here https://sourcegraph.com/github.com/Yelp/clusterman@517097a48fd2430589135d2d3b4c3a7b312353d1/-/blob/clusterman/autoscaler/pool_manager.py?L181 and assert that this client was never called ( which will confirm that we got an empty map for the nodes

TheNavigat: I would agree if all the other tests in the class didn't do the same thing .-. I wouldn't want this specific test to be out of place, so I need to either fix all of them, or keep it that way.

Do you think it's worth updating all of them?

There's also the fact that there's some contention over whether testing private methods in unit tests is acceptable. It definitely isn't in integration tests, but we might need a team discussion if we'll be avoiding testing private methods.

Buffer0x7cd: yeah , I agree testing internal functions are already part of the codebase so let's continue it for now.

Buffer0x7cd (approved):

PR #135, opened 2022-01-10
CLUSTERMAN-613: add default until for disable

BeryJu:
### Description

closes CLUSTERMAN-613

### Testing Done

working correctly.

### DIFF
--- a/clusterman/cli/toggle.py
+++ b/clusterman/cli/toggle.py
@@ -55,6 +55,10 @@ def disable(args: argparse.Namespace) -> None:
         "timestamp": {"N": str(int(time.time()))},
     }
 
+    if not args.until:
+        print("Default has changed; autoscaler will be re-enabled in 30m since no --until flag was used.")
+        args.until = "30m"
+
     if args.until:
         state["expiration_timestamp"] = {"N": str(parse_time_string(args.until).timestamp)}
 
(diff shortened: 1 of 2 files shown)


clusterman/cli/toggle.py
```diff
@@ -55,6 +55,10 @@ def disable(args: argparse.Namespace) -> None:
         "timestamp": {"N": str(int(time.time()))},
     }
 
+    if not args.until:
+        print("Default has changed; autoscaler will be re-enabled in 30m since no --untill flag was used.")
```
nemacysts: should we set this default closer to where the arg is defined and unconditionally print how long the autoscaler will be paused for?

nemacysts (approved):

clusterman/cli/toggle.py
nemacysts: and if we decide to keep this:
```suggestion
        print("Default has changed; autoscaler will be re-enabled in 30m since no --until flag was used.")
```

BeryJu: Initially I just changed Line 108 to add the default there, but that won't communicate the changed default to users, so I prefer this way, at least for a couple versions and until people have noticed the changed default.

Buffer0x7cd:
LGTM 🚢

Buffer0x7cd (approved):

PR #140, opened 2022-01-18
Drop Xenial support on GHA

TheNavigat:


nemacysts:
we'll also want to drop xenial from https://github.com/Yelp/clusterman/blob/master/jenkins.yaml#L10 since https://github.com/Yelp/clusterman/blob/fc4e52ab01b3b4134f5913af49bcd9671ad49651/package/itest/ubuntu.sh#L32 will hit the same broken path (or we could (/should) make sure that our itests use our internal apt repo even for our OS projects)

TheNavigat:
> we'll also want to drop xenial from https://github.com/Yelp/clusterman/blob/master/jenkins.yaml#L10 since
> 
> https://github.com/Yelp/clusterman/blob/fc4e52ab01b3b4134f5913af49bcd9671ad49651/package/itest/ubuntu.sh#L32
> 
> will hit the same broken path (or we could (/should) make sure that our itests use our internal apt repo even for our OS projects)

That's to fix our internal build. Let's fix GHAs first and then turn to internal builds.

nemacysts (approved):

PR #145, opened 2022-03-10
handle exception for non-exist instance_id

ilkinmammadzada:
### Description

Clusterman gets stuck getting messages from queue with non-exist instance_id.
The exception occurs and the message is returned to the queue, and this continues indefinitely.

### Testing Done

Test will be added for non-exist instance_id case

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -23,6 +23,7 @@ from typing import Sequence
 from typing import Type
 
 import arrow
+import botocore.exceptions
 import colorlog
 import staticconf
 
@@ -291,7 +292,11 @@ class DrainingClient:
 
 
 def host_from_instance_id(sender: str, receipt_handle: str, instance_id: str,) -> Optional[Host]:
-    instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    try:
+        instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    except botocore.exceptions.ClientError as e:
+        logger.warning(f"Couldn't describe instance: {e}")
+        return None
     if not instance_data:
         logger.warning(f"No instance data found for {instance_id}")
         return None
--- a/tests/draining/queue_test.py
+++ b/tests/draining/queue_test.py
@@ -14,6 +14,7 @@
 import socket
 
 import arrow
+import botocore.exceptions
 import mock
 import pytest
 import staticconf.testing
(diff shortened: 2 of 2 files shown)


ilkinmammadzada:
Jothi (I didnt find your account to tag :) ),
I think with this solution, process will delay for 15 seconds for no reason. 
Because client.py -> ec_describe_instances method has retry mechanism for ClientError, it will not throw exception to client code before three times retry.


`@retry(exceptions=botocore.exceptions.ClientError, tries=3, delay=5)
def ec2_describe_instances(instance_ids: Sequence[str]) -> List[InstanceDict]:`

ilkinmammadzada:
@jothi7891 Could please check my comment above?

jothi7891:
@ilkinmammadzada - i think that is acceptable since the `ClientError` can be thrown by`boto3` for other reasons as well hence the retry and delay has been embedded in to it.

jothi7891:
@ilkinmammadzada - appears like the tests are failing because of the ordering of python imports.

ilkinmammadzada:
> @ilkinmammadzada - appears like the tests are failing because of the ordering of python imports.

clusterman/draining/queue.py
```diff
@@ -291,7 +292,11 @@ def process_warning_queue(self) -> None:
 
 
 def host_from_instance_id(sender: str, receipt_handle: str, instance_id: str,) -> Optional[Host]:
-    instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    try:
+        instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    except botocore.exceptions.ClientError as e:
+        logger.warning(f"Couldn't describe instance: {e}")
```
nemacysts, line 298: could also use `logger.exception` and python will automatically include the traceback after logging the message :)

tests/draining/queue_test.py
```diff
@@ -479,6 +480,9 @@ def test_host_from_instance_id():
         mock_ec2_describe.return_value = [{"InstanceId": "i-123"}]
         assert host_from_instance_id(sender="aws", receipt_handle="rcpt", instance_id="i-123",) is None
 
+        mock_ec2_describe.side_effect = botocore.exceptions.ClientError({}, "")
```
nemacysts, line 483: might be nice to add a comment here like the one above describe the case we're testing here

PR #146, opened 2022-03-15
Handling exception while trying describe non-exist instance_id

ilkinmammadzada:
### Description

Exception occurs when try to describe instance to get host information with non-exist instance_id.
Message returns back to queue and process continuous indefinitely.

### Testing Done

Test case was added to check specific scenario.
Test case failed without my change and it was expected.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -23,7 +23,7 @@ from typing import Sequence
 from typing import Type
 
 import arrow
-import botocore.exceptions
+from botocore.exceptions import ClientError
 import colorlog
 import staticconf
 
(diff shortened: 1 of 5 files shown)


ilkinmammadzada:
@jothi7891 did you move strtobool  method for testing?

jothi7891:
@ilkinmammadzada - we decided to vendor the function as that library would soon be deprecated

ilkinmammadzada:
> @ilkinmammadzada - we decided to vendor the function as that library would soon be deprecated

@jothi7891 , it is good idea. 
Additionally, I moved method to more general util file.

clusterman/draining/queue.py
```diff
@@ -291,7 +292,11 @@ def process_warning_queue(self) -> None:
 
 
 def host_from_instance_id(sender: str, receipt_handle: str, instance_id: str,) -> Optional[Host]:
-    instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    try:
+        instance_data = ec2_describe_instances(instance_ids=[instance_id])
+    except ClientError as e:
+        logger.warning(f"Couldn't describe instance: {e}")
```
jothi7891: @ilkinmammadzada - can we get this as `logger.exception` to include the traceback ?

jothi7891 (approved):

PR #149, opened 2022-03-25
Reserve capacity was added to target capacity to avoid stuck

ilkinmammadzada:
### Description

Autoscaling component stuck while calculating target capacity in the following case:
If the resource request and the target capacity are non-zero, but the nodes haven't joined
the cluster yet, we just need to wait until they join before doing anything else

### Testing

We dont have test scenario for this case and _compute_target_capacity method.
I will try to add test cases as well in this PR.

### DIFF
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -308,7 +308,8 @@ class Autoscaler:
                 "Non-orphan fulfilled capacity is 0 and current target capacity > 0, not changing target to let the "
                 "new instances join"
             )
-            return current_target_capacity
+            reserve_capacity = self._get_reserve_capacity()
+            return current_target_capacity + reserve_capacity
 
         # If we get here, everything is non-zero and we can use the "normal" logic to determine scaling
         (most_constrained_resource, usage_pct,) = self._get_most_constrained_resource_for_request(
@@ -447,3 +448,7 @@ class Autoscaler:
             latest_non_zero_values[-1][0],
             sum([float(val) for __, val in latest_non_zero_values]) / len(latest_non_zero_values),
         )
+
+    def _get_reserve_capacity(self) -> float:
+        # logic will be implemented
+        return 0
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -276,7 +276,8 @@ class Autoscaler:
         #       this adds an extra autoscaling cycle before you can get all your resources.
         #
         # 4. If the resource request and the target capacity are non-zero, but the nodes haven't joined
-        #    the cluster yet, we just need to wait until they join before doing anything else.
+        #    the cluster yet, we just need to add one additional resource.
+        #    Because non-joined instance may not initialize and this may cause to stuck indefinitely.
 
         if all(requested_quantity is None for requested_quantity in resource_request):
             logger.info("No data from signal, not changing capacity")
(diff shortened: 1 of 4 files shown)


ilkinmammadzada:
@Buffer0x7cd , We should increase capacity for our specific condition as We agreed.
Also We thought that limitation logic should be implemented to avoid  unnecessary resource consumption.

But after little bit investigation I saw that We have limitation logic in the pool:

````
scaling_limits:
    min_capacity: 1
    max_capacity: 800
    max_weight_to_add: 100
    max_weight_to_remove: 100
    max_tasks_to_kill: 100
````

As I understood autoscaler doesn't create instances more than scaling_limits.max_capacity.

Long story short we have two way to solve issue:
1) Increase current_target_capacity by one, and rely on the pool scaling_limits logic. But this may cause unnecessary resource consumption (up to max capacity)

2) Implement limit logic inside of autoscaling. We can add new config (for ex: reserve_resource_max_limit_ratio) and calculate  net limit count with target capacity and this config.

````
autoscaling:
    default_signal_role: 'clusterman'
    setpoint: 0.7
    target_capacity_margin: 0.05   
````

What do you think? Which way seems more appropriate?

File: clusterman/autoscaler/autoscaler.py [hunk 1]
```diff
@@ -304,9 +305,9 @@ def _compute_target_capacity(self, resource_request: SignalResourceRequest) -> f
         elif non_orphan_fulfilled_capacity == 0:
             # Entering the main body of this method with non_orphan_fulfilled_capacity = 0 guarantees that
             # new_target_capacity will be 0, which we do not want (since the resource request is non-zero)
+            current_target_capacity += 1
```
nemacysts, line 308: are we ok with potentially over-scaling to start with? there are some pools that run the autoscaler on a period < how long it takes nodes to come online - this change would lead us to keep adding more capacity until that first node joins the cluster, right? (even if everything is ok)

ilkinmammadzada: Great point, I didn't know about frequency of calling method. We need to go with other way. Thank you @nemacysts !

ankit864: I agree with @nemacysts  current logic  may cause issues for some cluster.
I am thinking,  instead of this approach can we have timeout  which can recycle orphan node  if they are not joining cluster within some time limit as it will prevent us over-scaling also solves problem for waiting orphan node indefinitely.  thoughts ?

ilkinmammadzada: @ankit864 Cluster will not start running any workload while this "timeout".
Yes, waiting for "timeout" is better than "indefinitely". But is it acceptable?
Do we know approximately "instance joining time" ?

Buffer0x7cd: @ilkinmammadzada  I think the approach @ankit864 mentioned seems to be better suited since that would let us solve a few different issues regarding the ORPHAN node. The Key method to solve this problem is to force clusterman to launch new nodes even if nodes are stuck in the ORPHAN state. While Increasing the current_capacity can help us achieve this, we can achieve the same by deleting the nodes that are stuck in the ORPHAN state for a long time ( Deleting nodes would call a mismatch in target capacity and full-field capacity which would force clusterman to launch new nodes).

ilkinmammadzada: @Buffer0x7cd got your point.
Should we apply same "timeout" period to all pool or one config per pool?
It looks like some kind of "background process" to check ORPHAN nodes periodically.
I need to check code to find suitable place for this.

Buffer0x7cd: I think we can use a default timeout ( around 30 mins). As a starting point, I think we can use the same config for all pools ( since 30 mins are plenty of time to join a cluster)

PR #150, opened 2022-04-05
Terminating expired orphan resources in the autoscaler

ilkinmammadzada:
### Description

Autoscaling component stuck while calculating target capacity in the following case:
If the resource request and the target capacity are non-zero, but the nodes haven't joined
the cluster yet, we just need to wait until they join before doing anything else.
In this PR we will add functionality to clean up expired orphan resources.

### Testing Done

We don't have test scenario for this case.
I will try to add test cases as well in this PR.


clusterman/autoscaler/autoscaler.py
```diff
@@ -174,6 +174,7 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             self.target_capacity_gauge.set(new_target_capacity, {"dry_run": dry_run})
             self._emit_requested_resource_metrics(resource_request, dry_run=dry_run)
 
+        self.pool_manager.terminate_expired_orphan_instances()
```
jothi7891: i think we should be passing the `dry_run` here

jothi7891 (commented):
Need test cases but otherwise looks good

jothi7891:
lgtm

ilkinmammadzada:
> Need test cases but otherwise looks good

Test case was added for filtering, and I will try to add terminating method as well.

tests/autoscaler/pool_manager_test.py
```diff
@@ -553,3 +554,21 @@ def test_instance_kill_order(mock_pool_manager):
...
+            _make_metadata("sfr-0", "i-1", agent_state=AgentState.RUNNING, uptime=1500),
+            _make_metadata("sfr-1", "i-2", agent_state=AgentState.ORPHANED, uptime=500),
+            _make_metadata("sfr-1", "i-3", agent_state=AgentState.ORPHANED, uptime=2500),
+            _make_metadata("sfr-2", "i-4", agent_state=AgentState.RUNNING, uptime=2500),
+            _make_metadata("sfr-2", "i-5", agent_state=AgentState.RUNNING, uptime=2500),
+        ]
+    )
+
+    result = mock_pool_manager._get_expired_orphan_instances()
+
```
Buffer0x7cd, line 573: Let's avoid testing internal methods in tests, We have done this quite a lot in existing test case which we needs to be refactord, but we can avoid adding tests that are dependent on testing internal method. We can make _get_expired_orphan_instances a normal method instead of a internal method

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +205,41 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+
+        if dry_run:
+            return
+
+        for group_id in expired_orphan_instances:
+            self.resource_groups[group_id].terminate_instances_by_id(expired_orphan_instances[group_id])
+
+    def _get_expired_orphan_instances(self) -> Dict[str, List[str]]:
+        groups = self.resource_groups.values()
+        node_metadatas = self.get_node_metadatas()
```
Buffer0x7cd: It might be a better idea to use the state filter here so that we only query Running instances, since those are the only once we are intersted in, Example: https://github.com/Yelp/clusterman/blob/a8469f29b1c0f7059d116bb9dccde0b64e8764fe/clusterman/autoscaler/pool_manager.py#L483

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +205,41 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+            for metadata in node_metadatas:
+                if metadata.instance.group_id == group.id and self._is_expired_orphan_instance(metadata):
+                    expired_orphan_instances[group.id].append(metadata.instance.instance_id)
+
+        return expired_orphan_instances
+
+    def _is_expired_orphan_instance(self, metadata: ClusterNodeMetadata) -> bool:
+        if (
+            metadata.agent.state == AgentState.ORPHANED
+            and metadata.instance.uptime > 1800
```
Buffer0x7cd: Minor nit: Move the uptime value to config.py instead of hardcoding the value here. Alternatively make this method accept a timeout parameter with some default value.

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +205,41 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+            and metadata.instance.state == 'running'
```
Buffer0x7cd: We can use the state describe here (https://github.com/Yelp/clusterman/blob/a8469f29b1c0f7059d116bb9dccde0b64e8764fe/clusterman/autoscaler/pool_manager.py#L50) instead of explicitly mentioning it. This will keep the state tracking tidy

clusterman/autoscaler/autoscaler.py
```diff
@@ -174,6 +174,7 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             self.target_capacity_gauge.set(new_target_capacity, {"dry_run": dry_run})
             self._emit_requested_resource_metrics(resource_request, dry_run=dry_run)
 
+        self.pool_manager.terminate_expired_orphan_instances(dry_run=dry_run)
```
Buffer0x7cd: This method can fail due API limit restrictions or when we try to delete a large number of instances. I think we should add exception handling here and gracefully skip this step if we can’t terminate the instance ( We should be able to try terminating the instanct in next run).

we shouldn’t add this method in critical path of scaling since it deals with few exceptional case, so by wrapping in exception we can continue autoscaling the cluster in case of an error in method.

Buffer0x7cd: Also, It will be safer to enable this behind a feature flag ( using srv_config).

For example it can look something like

if recycle_orphan_enable:
    self.pool_manager.terminate_expired_orphan_instances

we can use the feature toggle to slowly test this in different pools.

ilkinmammadzada:
@Buffer0x7cd thanks a lot for all your valuable comment! I will fix them and ping you.

ilkinmammadzada:
@Buffer0x7cd I just finished requested changes, Could you please re-check?

clusterman/autoscaler/pool_manager.py
```diff
@@ -218,6 +218,38 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+
+        for group in groups:
+            for metadata in node_metadatas:
+                if metadata.instance.group_id == group.id and self._is_expired_orphan_instance(metadata, timeout):
+                    group_id_to_instance_ids[group.id].append(metadata.instance.instance_id)
+
+        return group_id_to_instance_ids
+
+    def _is_expired_orphan_instance(self, metadata: ClusterNodeMetadata, timeout: int) -> bool:
+        if metadata.agent.state == AgentState.ORPHANED and metadata.instance.uptime > timeout:
```
Buffer0x7cd: you could simplify it by using something like
`return (metadata.agent.state == AgentState.ORPHANED) and( metadata.instance.uptime > timeout)`

Buffer0x7cd: Looks good except this small change.

Buffer0x7cd:
Fix the tests and shipit :shipit:

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +203,35 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+        group_id_to_instance_ids = self.get_expired_orphan_instances()
+        logger.info(f"Terminating expired orphan instances: \n{group_id_to_instance_ids}")
+
+        if dry_run:
+            return
+
+        for group_id in group_id_to_instance_ids:
+            self.resource_groups[group_id].terminate_instances_by_id(group_id_to_instance_ids[group_id])
+
+    def get_expired_orphan_instances(self, timeout: int = 1800) -> Mapping[str, List[str]]:
```
nemacysts: nit:
```suggestion
    def get_expired_orphan_instances(self, timeout: int = 1800) -> Dict[str, List[str]]:
```

we generally prefer to type return values with more concrete types rather than with interfaces

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +203,35 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+        node_metadatas = self.get_node_metadatas(AWS_RUNNING_STATES)
+
+        group_id_to_instance_ids: Mapping[str, List[str]] = defaultdict(list)
+
+        for group in groups:
+            for metadata in node_metadatas:
+                if metadata.instance.group_id == group.id and self._is_expired_orphan_instance(metadata, timeout):
+                    group_id_to_instance_ids[group.id].append(metadata.instance.instance_id)
+
+        return group_id_to_instance_ids
```
nemacysts, line 232: might be nice to convert this to a dict so that future callers aren't surprised by this being a defaultdict

clusterman/autoscaler/pool_manager.py
```diff
@@ -203,6 +203,35 @@ def get_node_metadatas(self, state_filter: Optional[Collection[str]] = None) ->
...
+    def get_expired_orphan_instances(self, timeout: int = 1800) -> Mapping[str, List[str]]:
+        groups = self.resource_groups.values()
+        node_metadatas = self.get_node_metadatas(AWS_RUNNING_STATES)
+
+        group_id_to_instance_ids: Mapping[str, List[str]] = defaultdict(list)
+
+        for group in groups:
+            for metadata in node_metadatas:
+                if metadata.instance.group_id == group.id and self._is_expired_orphan_instance(metadata, timeout):
+                    group_id_to_instance_ids[group.id].append(metadata.instance.instance_id)
```
nemacysts: do we need to iterate over the groups? would something like 
```suggestion
        for metadata in node_metadatas:
            if self._is_expired_orphan_instance(metadata, timeout):
              
  group_id_to_instance_ids[metadata.instance.group_id].append(metadata.instance.instance_id)
```

if we want to make sure that the group id is one we care about we could make a set with all those ids and do something like `if metadata.instance.group_id in known_groups` though - instead of nesting loops

nemacysts (commented):
just have some nits - feel free to ignore me

ilkinmammadzada:
> just have some nits - feel free to ignore me

@nemacysts Thanks a lot for your valuable comments. All of them were solved, Could you please re-check?

nemacysts (approved):

PR #152, opened 2022-04-11
Use new clusterman_signals with Python 3.8

TheNavigat:
This fixes `make itest-external` and fixes the Clusterman bug we have seen when we rolled out today


BeryJu (approved):

TheNavigat:
Closing till we upgrade clusterman_signals

PR #156, opened 2022-05-05
Recycling orphan instances by default

ilkinmammadzada:
Description
We applied this feature to all pools in all clusters,
We are changing default value of setting to facilitate configuration and apply change to all future pools.

Testing Done
Test case was added already.


mattmb (approved):
Don't forget to check that this version deploys to prod successfully before you remove the srv-configs :-)

PR #157, opened 2022-05-12
Support draining warned Spot k8s instances

nemacysts:
We're currently doing nothing with the advance warning we get from AWS
for Spot instances in our K8s clusters.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -20,6 +20,7 @@ from typing import Dict
 from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
+from typing import Set
 from typing import Type
 
 import arrow
@@ -29,6 +30,7 @@ from botocore.exceptions import ClientError
 
 from clusterman.args import add_cluster_arg
 from clusterman.args import subparser
+from clusterman.aws.auto_scaling_resource_group import AutoScalingResourceGroup
 from clusterman.aws.aws_resource_group import AWSResourceGroup
 from clusterman.aws.client import ec2_describe_instances
 from clusterman.aws.client import sqs
@@ -38,8 +40,9 @@ from clusterman.aws.util import RESOURCE_GROUPS_REV
 from clusterman.config import load_cluster_pool_config
 from clusterman.config import POOL_NAMESPACE
 from clusterman.config import setup_config
+from clusterman.draining.kubernetes import drain as k8s_drain
 from clusterman.draining.mesos import down
-from clusterman.draining.mesos import drain
+from clusterman.draining.mesos import drain as mesos_drain
 from clusterman.draining.mesos import operator_api
 from clusterman.draining.mesos import up
 from clusterman.interfaces.resource_group import InstanceMetadata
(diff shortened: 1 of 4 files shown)


clusterman/draining/queue.py
```diff
@@ -248,7 +251,20 @@ def process_drain_queue(
...
+                        k8s_drain(
+                            kube_operator_client, host_to_process.hostname,
+                        )
+                    else:
+                        logger.info(
+                            f"Unable to drain {host_to_process.hostname} (no Kubernetes connector configured) - "
+                            "continuing to terminate anyway!",
+                        )
+                except Exception as e:
+                    logger.error(f"Failed to drain {host_to_process.hostname} continuing to terminate anyway: {e}")
```
ilkinmammadzada, line 265: We would like create something to churning/recycling instances to keep them fresh. 
CEP1869 - https://yelpwiki.yelpcorp.com/display/SEC/Project%3A+Kubernetes+Instance+Upgrade
We planned to protect applications with PDB. Draining should respects protection logic.
You used  pod eviction what is perfect choice for this. But we would like to clean-up pods fully before termination. Because application can be corrupted/worked with bad performance if we don't respect PDB. 
I think we should discuss this in #kubernetes-instance-upgrade, please ping me once you have availability :)

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -62,6 +65,46 @@ def __init__(self, cluster: str, pool: Optional[str]) -> None:
...
+                body=V1beta1Eviction(
+                    metadata=V1ObjectMeta(name=pod.metadata.name, namespace=pod.metadata.namespace,),
+                    delete_options=V1DeleteOptions(
+                        # we don't want to block on eviction as we're potentially evicting a ton of pods
+                        # AND there's a delay before we go ahead and terminate
+                        # AND at Yelp we run a script on shutdown that will also try to drain one final time.
+                        propagation_policy="Background",
+                    ),
+                ),
+            )
```
ilkinmammadzada, line 100: btw this method throws exception (ApiException - status:429, reason: too many requests) when you try to delete pod more than expected (PDB).

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -62,6 +65,46 @@ def __init__(self, cluster: str, pool: Optional[str]) -> None:
...
+        )
+
+    def _list_all_pods_on_node(self, hostname: str) -> List[KubernetesPod]:
+        return self._core_api.list_pod_for_all_namespaces(field_selector=f"spec.nodeName={hostname}")
+
+    def _evict_tasks_from_node(self, hostname: str) -> None:
+        pods_to_evict = [
+            pod
+            for pod in self._list_all_pods_on_node(hostname)
+            if not self._pod_belongs_to_daemonset(pod) and pod.status and pod.status.phase == "Running"
```
ilkinmammadzada, line 80: This will ignore pods which scheduled right before "cordon", Because they will be "Pending" phase for a while.

clusterman/draining/queue.py
```diff
@@ -248,7 +251,20 @@ def process_drain_queue(
                     self.submit_host_for_termination(host_to_process)
             elif host_to_process.scheduler == "kubernetes":
                 logger.info(f"Kubernetes host to drain and submit for termination: {host_to_process}")
-                self.submit_host_for_termination(host_to_process, delay=0)
+                try:
+                    if kube_operator_client:
+                        k8s_drain(
+                            kube_operator_client, host_to_process.hostname,
```
ilkinmammadzada, line 257: We can't drain k8s node with host_to_process.hostname. We need agent.agent_id (node name)
P.S. I know this PR is stale, just in case..

PR #159, opened 2022-06-02
CLUSTERMAN-637: Add basic Rookout support

BeryJu:
### Description

COMPINFRA-1009, adding it to clusterman first as there's a bunch of tooling and setup already done that makes it easier to add rookout to paasta services, compared to other projects

Rookout is disabled by default so this should be safe to deploy, can be enabled at any time from soaconfigs

### DIFF
--- a/clusterman/tools/rookout.py
+++ b/clusterman/tools/rookout.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from os import getenv
+
+
+def enable_rookout():
+    """Enable rookout if environment variables are set"""
+    if getenv("ROOKOUT_ENABLE", "") != "1":
+        return
+    import rook
+
+    rook.start(token=getenv("ROOKOUT_TOKEN"))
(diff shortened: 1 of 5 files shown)


requirements-minimal.txt
```diff
@@ -19,3 +19,7 @@ signalfx
 simplejson
 sortedcontainers
 typing-extensions
+rook==0.1.160
+distro==1.7.0
+funcsigs==1.0.2
+psutil==5.9.1
```
nemacysts:
```suggestion
rook
```

this file should only include libraries that are directly imported in production code

nemacysts (approved):
do we need to do anything to enable rookout for our custom signals? or will they automatically be instrumented by this?

nemacysts (changes requested):
err, misclicked

BeryJu:
> do we need to do anything to enable rookout for our custom signals? or will they automatically be instrumented by this?

Good question, since they are completely independent sub-processes (afaik that's the case), we'll have to import the SDK there too. Also, I can't remember if the signals are continuously running, or if they are started, do their thing and then exit, because if the latter is the case then rookout doesn't work that well with that yet

clusterman/tools/rookout.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from os import getenv
+
+
+def enable_rookout():
```
nemacysts:
```suggestion
def enable_rookout() -> None:
```

dunno how i missed that in the original review :p

nemacysts (approved):

clusterman/tools/rookout.py
BeryJu: do'h me neither

nemacysts (approved):

BeryJu:
I'm a bit confused why all the tests succeed since both of the clusterman imports were incorrect (`from clusterman.clusterman.tools.rookout import enable_rookout` instead of `from clusterman.tools.rookout import enable_rookout`)

PR #164, opened 2022-06-24
Avoid pods scheduling after marking for termination

ilkinmammadzada:
### Description

1) Clusterman get pods count in node and then start termination process.
2) Kube scheduler schedule new pods before Clusterman terminating node
https://jira.yelpcorp.com/browse/CLUSTERMAN-642

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -444,6 +444,7 @@ class PoolManager:
             self.cluster_connector.freeze_agent(node_metadata.agent)
 
             task_count_realtime = self.cluster_connector.get_task_count_realtime(node_metadata.agent)
+            #node_metadata.agent.task_count = task_count_realtime not sure yet
             if killed_task_count + task_count_realtime > self.max_tasks_to_kill:  # case 4
                 logger.info(
                     f"Killing instance {instance_id} with {task_count_realtime} tasks (realtime) would take us "
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -439,7 +439,7 @@ class PoolManager:
                     continue
 
             logger.info(f"freezing {instance_id} for termination")
-            self.cluster_connector.freeze_agent(node_metadata.agent)
+            self.cluster_connector.freeze_agent(node_metadata.agent.agent_id)
             logger.info(f"marking {instance_id} for termination")
             marked_nodes[group_id].append(node_metadata)
             rem_group_capacities[group_id] -= instance_weight
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -545,7 +545,7 @@ class PoolManager:
             node_metadata: ClusterNodeMetadata,
         ) -> Tuple[int, int, int, int, int, int]:
             return (
-                0 if node_metadata.agent.is_freezed else 1,
+                0 if node_metadata.agent.is_frozen else 1,
                 0 if node_metadata.agent.state == AgentState.ORPHANED else 1,
                 0 if node_metadata.agent.state == AgentState.IDLE else 1,
                 0 if node_metadata.instance.is_stale else 1,
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -423,8 +423,6 @@ class PoolManager:
                 logger.info(f"Resource group {group_id} is at target capacity; skipping {instance_id}")
                 continue
 
-            task_count_realtime = self.cluster_connector.get_task_count_realtime(node_metadata.agent)
-
             if (
                 killed_task_count + node_metadata.agent.task_count > self.max_tasks_to_kill or
                 killed_task_count + self.cluster_connector.get_task_count_realtime(
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -423,7 +423,8 @@ class PoolManager:
                 logger.info(f"Resource group {group_id} is at target capacity; skipping {instance_id}")
                 continue
 
-            if killed_task_count + node_metadata.agent.task_count > self.max_tasks_to_kill:  # case 2
+            task_count_realtime = self.cluster_connector.get_task_count_realtime(node_metadata.agent)
+            if killed_task_count + task_count_realtime > self.max_tasks_to_kill:  # case 2
                 logger.info(
                     f"Killing instance {instance_id} with {node_metadata.agent.task_count} tasks would take us "
                     f"over our max_tasks_to_kill of {self.max_tasks_to_kill}. Skipping this instance."
@@ -439,6 +440,7 @@ class PoolManager:
                     continue
 
             logger.info(f"marking {instance_id} for termination")
+            #TODO taint node - noschedule ---
             marked_nodes[group_id].append(node_metadata)
             rem_group_capacities[group_id] -= instance_weight
             curr_capacity -= instance_weight
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -182,7 +182,7 @@ class PoolManager:
         :param dry_run: if True, do not modify the state of the cluster, just log actions
         """
 
-        marked_nodes_by_group = self._choose_nodes_to_prune(new_target_capacity, group_targets)
+        marked_nodes_by_group = self._choose_nodes_to_prune(new_target_capacity, group_targets, dry_run)
 
         if not dry_run:
             if self.draining_enabled:
@@ -352,6 +352,7 @@ class PoolManager:
         self,
         new_target_capacity: float,
         group_targets: Optional[Mapping[str, float]],
+        dry_run: bool = False,
     ) -> Mapping[str, List[ClusterNodeMetadata]]:
         """Choose nodes to kill in order to decrease the capacity on the cluster.
 
@@ -439,7 +440,8 @@ class PoolManager:
                     continue
 
             logger.info(f"freezing {instance_id} for termination")
-            self.cluster_connector.freeze_agent(node_metadata.agent.agent_id)
+            if not dry_run:
+                self.cluster_connector.freeze_agent(node_metadata.agent.agent_id)
             logger.info(f"marking {instance_id} for termination")
             marked_nodes[group_id].append(node_metadata)
             rem_group_capacities[group_id] -= instance_weight
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -452,6 +452,7 @@ class PoolManager:
                         f"Killing instance {instance_id} with {task_count_realtime} tasks  would take us "
                         f"over our max_tasks_per_node_to_kill of {self.max_tasks_per_node_to_kill}. Skipping this instance."
                     )
+                    # not sure if we need unfreeze or not, maybe it should be terminated in next try.
                     logger.info(f"unfreezing {instance_id} for termination")
                     self.cluster_connector.unfreeze_agent(node_metadata.agent)
                     continue
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -405,6 +405,8 @@ class PoolManager:
             #  1) The resource group the node belongs to can't be reduced further.
             #  2) Killing the node's tasks would take over the maximum number of tasks we are willing to kill.
             #  3) Killing the node would bring us under our target_capacity of non-orphaned nodes.
+            #  4) agent.task_count may be outdated date (case 3), and pods may be terminated unintentionally.
+            #     Therefore, double-checking task count from kube-api to avoid that case.
             # In each of the cases, the node has been removed from consideration and we jump to the next iteration.
 
             instance_id = node_metadata.instance.instance_id
@@ -423,12 +425,7 @@ class PoolManager:
                 logger.info(f"Resource group {group_id} is at target capacity; skipping {instance_id}")
                 continue
 
-            if (
-                killed_task_count + node_metadata.agent.task_count > self.max_tasks_to_kill or
-                killed_task_count + self.cluster_connector.get_task_count_realtime(
-                    node_metadata.agent) > self.max_tasks_to_kill
-            ):  # case 2
-                # fix log to write correct count
+            if killed_task_count + node_metadata.agent.task_count > self.max_tasks_to_kill:  # case 2
                 logger.info(
                     f"Killing instance {instance_id} with {node_metadata.agent.task_count} tasks would take us "
                     f"over our max_tasks_to_kill of {self.max_tasks_to_kill}. Skipping this instance."
@@ -442,6 +439,14 @@ class PoolManager:
                         f"our target_capacity for non-orphan boxes. Skipping this instance."
                     )
                     continue
+
+            task_count_realtime = self.cluster_connector.get_task_count_realtime(node_metadata.agent)
+            if killed_task_count + task_count_realtime > self.max_tasks_to_kill:  # case 4
+                logger.info(
+                    f"Killing instance {instance_id} with {task_count_realtime} tasks (realtime) would take us "
+                    f"over our max_tasks_to_kill of {self.max_tasks_to_kill}. Skipping this instance."
+                )
+                continue
 
             logger.info(f"marking {instance_id} for termination")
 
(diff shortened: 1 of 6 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -241,6 +241,10 @@ def _count_batch_tasks(self, node_ip: str) -> int:
                     break
         return count
 
+    def _freeze_agent(self, agent_id: str) -> None:
+        body = {"spec": {"unschedulable": True}}
+        self._core_api.patch_node(agent_id, body)
```
BeryJu: it would be nice to have some tests for this but probably not easily doable

BeryJu (approved):
LGTM

clusterman/interfaces/cluster_connector.py
```diff
@@ -46,6 +46,14 @@ def get_agent_metadata(self, ip_address: Optional[str]) -> AgentMetadata:
             return AgentMetadata()
         return self._get_agent_metadata(ip_address)
 
+    def freeze_agent(self, agent_id: Optional[str]) -> None:
```
Rob-Johnson: why do we need a wrapper method here? if `agent_id` is always required, can we just make it non optional?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -241,6 +241,10 @@ def _count_batch_tasks(self, node_ip: str) -> int:
                     break
         return count
 
+    def _freeze_agent(self, agent_id: str) -> None:
+        body = {"spec": {"unschedulable": True}}
```
Rob-Johnson: how does this compare to using a `NoSchedule` taint? if we set a `NoSchedule` taint with a descriptive name, I think it'll make it easier to determine why a node is unschedulable?

ilkinmammadzada: it also add NoSchedule taint automatically. "kubectl cordon $node" aldo does same, I found it from kubectl with -v9

Rob-Johnson: in that case, can we just create the taint directly (with a more descriptive name than unschedulable)?

ilkinmammadzada: Sure, But how can we achieve stopping new pods scheduling ?
I "cordon" node when we mark node for termination. it will avoid new pods scheduling before real termination.
The gap between "marking for termination" and "real termination" is more than 30 seconds for jenkins pool. 
And kube-scheduler schedules new pods (jenkins job) in this gap, and we delete pods right before real termination (node). deletion of these pods causes issue for someone's pipeline.

ilkinmammadzada: I forgot to mention that "unschedulable" wasn't selected by me. It is field of NodeSpec: 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#nodespec-v1-core
https://kubernetes.io/docs/concepts/architecture/nodes/#manual-node-administration

Rob-Johnson: `cordon` does is set `node.Spec.Unschedulable` to be true (https://github.com/kubernetes/kubectl/blob/master/pkg/drain/cordon.go#L89). as you say, this stops pods from being scheduled on a host.
when launching a new pod, the scheduler also checks whether pods have tolerations for all the taints on the host. taints can have two effects - `NoSchedule` or `NoExecute`. both have the impact of preventing any pods being scheduled unless they have a matching toleration, but NoExecute also removes running pods.

So they both have the same effect for us - stopping pods from being scheduled on a host (though, if we use a taint, an operator *could* circumvent it by setting the appropriate taint on a pod. unlikely to be a problem for us).

the reason I'm advocating for a taint instead of just marking a node as `unschedulable` is that the taint gives us the opportunity to put a descriptive reason on *why* we're stopping pods being launched. a node might be unschedulable for a bunch of reasons - it'd be good if we could avoid adding another reason (clusterman is shutting the node down) into that mix.

ilkinmammadzada: I got your point! it is good idea. I will taint node with something like "ClustermanScaleDown=true:NoSchedule". Actually I will try to find better name :) thank you!

Rob-Johnson: sounds good.
can you make sure you add a namespaced prefix for the taint - f.e. something like `clusterman.yelp.com/terminating`?

clusterman/mesos/mesos_cluster_connector.py
```diff
@@ -126,3 +126,6 @@ def _is_batch_framework(self, framework_name: str) -> bool:
         """If the framework matches any of the prefixes in self.non_batch_framework_prefixes
         this will return False, otherwise we assume the task to be a batch task"""
         return not any([framework_name.startswith(prefix) for prefix in self.non_batch_framework_prefixes])
+
+    def _freeze_agent(self, agent_id: str) -> None:
+        return
```
nemacysts: i'd maybe add a comment here about how we could put the box in question into maintenance mode, but since we're almost done getting rid of mesos we're not putting in the effort to do so

clusterman/autoscaler/pool_manager.py
```diff
@@ -438,6 +438,8 @@ def _choose_nodes_to_prune(
                     )
                     continue
 
+            logger.info(f"freezing {instance_id} for termination")
+            self.cluster_connector.freeze_agent(node_metadata.agent.agent_id)
```
nemacysts: i'm still a little worried about racing against this taint now - in the time we get all nodes and iterate through them, we could be tainting a box that had something scheduled onto it (and ending up in the same situation where a box gets killed and was running something that doesn't tolerate interruptions)

clusterman/interfaces/cluster_connector.py
```diff
@@ -90,6 +98,10 @@ def get_cluster_total_resources(self) -> ClustermanResources:
     def _get_agent_metadata(self, ip_address: str) -> AgentMetadata:  # pragma: no cover
         pass
 
+    @abstractmethod
+    def _freeze_agent(self, agent_id: str) -> None:
+        pass
```
nemacysts: it feels a little weird to have a private abstract method (i.e., a private function that needs to be implemented as part of the public interface for people subclassing this) - maybe if we get rid of the wrapper above we can make this public with the same name?

additionally, thoughts on renaming this as cordon/uncordon? mesos is not long for this world at yelp and we might as well use standard/consistent terminology

clusterman/autoscaler/pool_manager.py
nemacysts: also, do we want to do anything about a node getting cordoned and clusterman crashing/getting killed/etc before terminating the cordoned box?

nemacysts (commented):
so sorry! i never noticed that i never published my review on your first iteration of this pr!

clusterman/autoscaler/pool_manager.py
ilkinmammadzada: For first question: It is part of main change which totally solve race condition. I am just applying change to production with small iteration. + I will check result after this change and I will solve race condition in case of necessity (because it is costy operation)

clusterman/interfaces/cluster_connector.py
ilkinmammadzada: I am just removing wrapper method, "cordon" (is_cordoned) has other meaning in clusterman, so I wouldn't use it :) . We add this tag to AWS side, and then check and decide whether it is "is_cordoned"

clusterman/interfaces/cluster_connector.py
ilkinmammadzada: Thanks! removed

clusterman/autoscaler/pool_manager.py
ilkinmammadzada: For second question: there are two possibilities:
1) crash after draining_client.submit_instance_for_draining: We don't need to think about this case
2) crash before draining_client.submit_instance_for_draining: In this case clusterman will select same node as victim due to our sort logic (_prioritize_killable_nodes). But there is still a possibility (small) that other nodes (for ex: newly created nodes which have less tasks) are ahead of the line. We will have "cordoned (freezed)" node for short (hope) period. 

Thanks for mentioned it. 
I will dig a little bit.

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Sure, Added, thanks!

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -128,6 +129,17 @@ def get_unschedulable_pods(
                 unschedulable_pods.append((pod, self._get_pod_unschedulable_reason(pod)))
         return unschedulable_pods
 
+    def freeze_agent(self, agent_id: str) -> None:
+        try:
+            body = {
+                "spec": {
+                    "taints": [{"effect": "NoSchedule", "key": "clusterman.yelp.com/terminating", "value": "True"}]
```
Rob-Johnson: ooh, sorry, final thing - it might be useful to set the value of this taint to a unix timestamp rather than just `True`? that way we can also see how long its been in that state?

Rob-Johnson (approved):
fix and ship 🚢 

can you also please rebase your commits into one or two smaller ones? it's hard to understand the current commit log

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: @Rob-Johnson  Yeah sure, it will be useful.

I tested it, kube-scheduler doesn't schedule new pods to tainted node as expected.
But kubectl command doesn't show "SchedulingDisabled" in status section. is it ok?
I think it is useful information. What do you think about change spec.unschedulable as well?

ilkinmammadzada: @Rob-Johnson  BTW Taint has timeAdded field. I will add it there.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#taint-v1-core

nemacysts: it might be nicer for humans to have it be the taint value so that it's more visible (the docs also mention `TimeAdded represents the time at which the taint was added. It is only written for NoExecute taints` which makes me a little worried that the field isn't writable for `NoSchedule` taints)

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -128,6 +129,17 @@ def get_unschedulable_pods(
...
+                }
+            }
+            self._core_api.patch_node(agent_id, body)
+        except ApiException as e:
+            logger.warning(f"Failed to patch {agent_id}: {e}")
```
nemacysts:
```suggestion
            logger.warning(f"Failed to freeze {agent_id}: {e}")
```

clusterman/interfaces/cluster_connector.py
nemacysts: unfortunate naming, i guess :p - now we're differing from kubernetes terminology 

oh well!

clusterman/autoscaler/pool_manager.py
nemacysts: i'm not 100% sold that *just* cordoning termination candidates will permanently fix the issue we're seeing here, but i'm willing to try this out

clusterman/mesos/mesos_cluster_connector.py
```diff
@@ -73,6 +73,9 @@ def get_resource_allocation(self, resource_name: str) -> float:
     def get_resource_total(self, resource_name: str) -> float:
         return sum(getattr(total_agent_resources(agent), resource_name) for agent in self._agents_by_ip.values())
 
+    def freeze_agent(self, agent_id: str) -> None:
+        return
```
nemacysts, line 78: https://github.com/Yelp/clusterman/pull/164#discussion_r914209829

clusterman/autoscaler/pool_manager.py
ilkinmammadzada: I checked many cases (+Cases in the ticket), ~all of them were scheduled after "marking for termination".  Yeah, let's try, see result. We have ready dashboard to check that, I will fix it in case of necessity.

clusterman/kubernetes/kubernetes_cluster_connector.py
Rob-Johnson: > @Rob-Johnson Yeah sure, it will be useful.
> 
> I tested it, kube-scheduler doesn't schedule new pods to tainted node as expected. But kubectl command doesn't show "SchedulingDisabled" in status section. is it ok? I think it is useful information. What do you think about change spec.unschedulable as well?

I think that's fine - the taint is 'enough' to stop things being scheduled

Rob-Johnson: > it might be nicer for humans to have it be the taint value so that it's more visible 

agreed - if we can do both, then great

clusterman/mesos/mesos_cluster_connector.py
ilkinmammadzada: AFAIK we use is_cordoned:true to aws instance to maintain purpose also. So clusterman marks it not_killable: pool_manager -> _is_node_killable

clusterman/autoscaler/pool_manager.py
ilkinmammadzada: @nemacysts , I implemented "is_freezed" logic and added it to sort_key with the highest priority. So Clusterman will select this node for termination. What do you think? I'm still thinking about if I missed any edge cases.

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: timestamp was added to the value.

clusterman/interfaces/types.py
```diff
@@ -20,6 +20,7 @@ class AgentMetadata(NamedTuple):
     allocated_resources: ClustermanResources = ClustermanResources()
     batch_task_count: int = 0
     is_safe_to_kill: bool = True
+    is_freezed: bool = False
```
nemacysts: super-nit:
```suggestion
    is_frozen: bool = False
```

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -184,12 +197,23 @@ def _get_agent_metadata(self, node_ip: str) -> AgentMetadata:
...
             allocated_resources=allocated_node_resources(self._pods_by_ip[node_ip]),
             is_safe_to_kill=self._is_node_safe_to_kill(node_ip),
+            is_freezed=self._is_node_freezed(node),
             batch_task_count=self._count_batch_tasks(node_ip),
             state=(AgentState.RUNNING if self._pods_by_ip[node_ip] else AgentState.IDLE),
             task_count=len(self._pods_by_ip[node_ip]),
             total_resources=total_node_resources(node, self._excluded_pods_by_ip.get(node_ip, [])),
         )
 
+    def _is_node_freezed(self, node: KubernetesNode) -> bool:
```
nemacysts:
```suggestion
    def _is_node_frozen(self, node: KubernetesNode) -> bool:
```

clusterman/mesos/mesos_cluster_connector.py
nemacysts: I meant that we could add a comment here mentioning why we're not doing the equivalent action in mesos (which would be putting the agent in maintenance mode)

although, actually: it might be good to add a log statement or something here - otherwise, the logs for mesos autoscalers will incorrectly say `"freezing {instance_id} for termination"` if i'm reading things correctly

clusterman/autoscaler/pool_manager.py
nemacysts: yea, sounds good!

clusterman/mesos/mesos_cluster_connector.py
ilkinmammadzada: Hmm. I got your point. Should I implement same logic for mesos? I skipped it because I thought that we don't use mesos anymor

nemacysts: we still use mesos in a few cases (although these are going away slowly)

it's not worth adding the same logic here (imo), but we should probably log something so that the logs aren't confusing. right now they'll say that mesos instances are being frozen, but that's not actually true

worst case, a comment that reads something like `# intentionally not implemented as Mesos is deprecated at Yelp` would work just as well (but would require whoever gets confused by the misleading log message in a mesos autoscaler to read the code :p)

nemacysts (approved):

PR #173, opened 2022-07-08
Pending phase pods are splitted

ilkinmammadzada:
### Description

CLUSTERMAN-642
We use only "Running" pods for all considerations like: task_count (node), resource allocation (pool, node), definition of state (node), and etc.

We should consider Pending pods which already assigned to node, Because this pods will running asap.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -75,9 +75,9 @@ class KubernetesClusterConnector(ClusterConnector):
         self._nodes_by_ip = self._get_nodes_by_ip()
         (
             self._pods_by_ip,
-            self._pending_pods,
+            self._unschedulable_pending_pods,
             self._excluded_pods_by_ip,
-        ) = self._get_pods_by_ip_or_pending()
+        ) = self._get_pods_information()
 
     def get_num_removed_nodes_before_last_reload(self) -> int:
         previous_nodes = self._prev_nodes_by_ip
@@ -118,17 +118,8 @@ class KubernetesClusterConnector(ClusterConnector):
         self,
     ) -> List[Tuple[KubernetesPod, PodUnschedulableReason]]:
         unschedulable_pods = []
-        for pod in self._pending_pods:
-            is_unschedulable = False
-            if not pod.status or not pod.status.conditions:
-                logger.info("No conditions in pod status, skipping")
-                continue
-
-            for condition in pod.status.conditions:
-                if condition.reason == "Unschedulable":
-                    is_unschedulable = True
-            if is_unschedulable:
-                unschedulable_pods.append((pod, self._get_pod_unschedulable_reason(pod)))
+        for pod in self._unschedulable_pending_pods:
+            unschedulable_pods.append((pod, self._get_pod_unschedulable_reason(pod)))
         return unschedulable_pods
 
     def freeze_agent(self, agent_id: str) -> None:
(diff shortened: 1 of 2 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -265,6 +256,29 @@ def _count_batch_tasks(self, node_ip: str) -> int:
                     break
         return count
 
+    def _is_pod_scheduled(self, pod: KubernetesPod) -> bool:
```
nemacysts: this name worries me a tiny bit as i could see myself (or a future yelper) thinking that this could return `True` for a Pod that was scheduled and is currently running - should this also return `True` in that case?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -248,9 +239,9 @@ def _get_pods_by_ip_or_pending(
             if self._pod_belongs_to_pool(pod):
                 if exclude_daemonset_pods and self._pod_belongs_to_daemonset(pod):
                     excluded_pods_by_ip[pod.status.host_ip].append(pod)
-                elif pod.status.phase == "Running":
+                elif pod.status.phase == "Running" or self._is_pod_scheduled(pod):
                     pods_by_ip[pod.status.host_ip].append(pod)
-                else:
+                elif self._is_pod_unschedulable(pod):
```
nemacysts: do we need this extra method? if a Pod doesn't belong to a daemonset (in any phase) or if a Pod isn't running/about to be running, it can only be pending, right? or were we getting error/terminated/etc pods here and counting these as pending before?

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Yes, You are right, I named it "_is_pod_pending_scheduled" but I didn't like so much. Let me think again

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: it can be Succeeded and Failed for jobs. that's the main reason of phase check.
And "pending" phase doesn't mean that pod is unschedulable: `The Pod has been accepted by the Kubernetes cluster, but one or more of the containers has not been set up and made ready to run. This includes time a Pod spends waiting to be scheduled as well as the time spent downloading container images over the network`

If I understood correctly we only care about unschedulable pods, Because we use "_pending_pods" only for "get_unschedulable_pods" method. So pending_pods=unschedulable_pods.
What do you think?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -248,11 +239,11 @@ def _get_pods_by_ip_or_pending(
...
                 if exclude_daemonset_pods and self._pod_belongs_to_daemonset(pod):
                     excluded_pods_by_ip[pod.status.host_ip].append(pod)
-                elif pod.status.phase == "Running":
+                elif pod.status.phase == "Running" or self._is_recently_scheduled(pod):
                     pods_by_ip[pod.status.host_ip].append(pod)
-                else:
-                    pending_pods.append(pod)
-        return pods_by_ip, pending_pods, excluded_pods_by_ip
+                elif self._is_unschedulable(pod):
+                    unschedulable_pods.append(pod)
```
nemacysts, line 246: should we add back in an `else` and log the pods we didn't categorize?

nemacysts (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Done

PR #175, opened 2022-08-03
Kubernetes draining logic was implemented

ilkinmammadzada:
### Description

Please fill out!

### Testing Done

working correctly.


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -133,6 +134,62 @@ def freeze_agent(self, agent_id: str) -> None:
...
+            if not self._cordon_node(agent_id):
+                return False
+            logger.info(f"Evicting pods on {agent_id}...")
+            if not self._evict_tasks_from_node(agent_id):
+                return False
+            logger.info(f"Drained  {agent_id}")
+            return True
+        except Exception as e:
+            logger.warning(f"Failed to drain {agent_id}: {e}")
+            return False
```
piax93: I wonder if we should un-cordon the node if this fails half-ways, otherwise I feel like there may be the risk of ending up with a lot of "unschedulable" infra. Or maybe a revert after a certain number of failures.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -133,6 +134,62 @@ def freeze_agent(self, agent_id: str) -> None:
...
+                self._core_api.create_namespaced_pod_eviction(
+                    name=pod.metadata.name,
+                    namespace=pod.metadata.namespace,
+                    body=V1beta1Eviction(
+                        metadata=V1ObjectMeta(name=pod.metadata.name, namespace=pod.metadata.namespace, ),
+                        delete_options=V1DeleteOptions(
+                            # we don't want to block on eviction as we're potentially evicting a ton of pods
+                            # AND there's a delay before we go ahead and terminate
+                            # AND at Yelp we run a script on shutdown that will also try to drain one final time.
+                            propagation_policy="Background",
```
piax93, line 201: This should probably be set somewhere in a configuration

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Good Idea. revert after certain number of tries or minutes++
I am trying to find best way.

piax93 (commented):
Looks reasonable to me, there are just a few issues with pre-commit hooks and mypy annotations in CI tests. I'd also get some second opinion from someone in CI.

ilkinmammadzada:
@piax93 Sure, I have one more Idea for expired draining. We can create a setting to decide what action should be happen when threshold time expired. Uncordon node or force terminate instance (ignore PDB). 
I will return to this PR after creating PDBs in prod. Thanks for review! :)

clusterman/draining/queue.py
```diff
@@ -357,6 +390,20 @@ def host_from_instance_id(
...
         logger.warning(f"Couldn't derive hostname from IP via DNS for {ip}")
         return None
+    try:
+        agent_id = instance_data[0]["PrivateDnsName"]
+    except KeyError:
+        logger.warning(f"No Private Dns Name found for {instance_id}")
+        return None
+    try:
+        for tag in instance_data[0]["Tags"]:
+            if "puppet:role::" in tag["Key"]:
```
piax93: Really don't know what'd be the best approach here, cause I feel like entangling things with Puppet parameters is not the best thing, but at the same time wouldn't know if there's a much better way.
Would throwing the private DNS name, which should correspond to the k8s node ID, to the k8s APIs and then read the pool label out of that be much worse?

ilkinmammadzada: I totally agree with you! I checked both and I was not happy none of them :) 
I think best way is adding new tag to the instance: https://sourcegraph.yelpcorp.com/misc/terraform-code/-/blob/modules/paasta_spot_asg_cluster/main.tf?subtree=true#L90:1

I am not sure whether adding new tag in the module will add tag to all existing instances.
It is still ok if it will not add, we can add it with small script. 

What do you think?

piax93: I'm pretty sure that to do that you'd have to update the tag specification on the launch template for the ASGs, which then would only affect the newly launched instances.

ilkinmammadzada: I am switching to DNS name now, then we can think about.

ilkinmammadzada: @piax93 probably you said `KeyName` value which is different for mesos and k8s

PR #177, opened 2022-08-22
Clusterman-582, Add label filter to get all pods for a pool

akshaysharma096:
## Description
The last PR - Yelp/clusterman#174 caused DAR-1678. The DAR was caused as there was not correct srv-config present in `stagef` and `devc`. This PR adds the feature of doing the server-side filtering of pods but that filtering can be switched on based on a feature flag called - `use_labels_for_pods`

PR for kubestage - 

### Why are we doing the change?
Currently Clusterman gets the all the pods while running `clusterman status`, which is very inefficient, this change is part of a big change where labels are added to different pod operators and clusterman can then filter the pods by labels for a pool rather than getting all pods at the same time.

Use label based filter to find out pods in a pool - https://github.com/Yelp/paasta/pull/3383

### How to release the change safely?
We have added a code snippet that will use `yelp.com/pool` for the nodes first and then go to `paasta.yelp.com/pool` if it does not exist.
```py
def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
        # TODO(https://jira.yelpcorp.com/browse/CLUSTERMAN-659)
        node_label_selector = self.pool_config.read_string(
            "node_label_key", default=self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
        )
        label_selector = f"{node_label_selector}={self.pool}"
        pool_nodes = self._core_api.list_node(label_selector=label_selector).items
        return {get_node_ip(node): node for node in pool_nodes}
```

This PR adds the feature of doing the server-side filtering of pods but that filtering can be switched on based on a feature flag called - `use_labels_for_pods`. 
To release the server side filtering feature for each cluster, we will make srv-config change and roll out gradually to each cluster and pool. 

```py
def reload_state(self) -> None:
        logger.info("Reloading nodes")

        self._core_api = CachedCoreV1Api(self.kubeconfig_path)

        # store the previous _nodes_by_ip for use in get_removed_nodes_before_last_reload()
        self._prev_nodes_by_ip = copy.deepcopy(self._nodes_by_ip)
        logger.info("Reloading pods")
        (self._pods_by_ip, self._unschedulable_pods, self._excluded_pods_by_ip,) = (
            self._get_pods_info_with_label()
            if self.pool_config.read_bool("use_labels_for_pods", default=False)
            else self._get_pods_info()
        )
```


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -215,14 +215,13 @@ def _is_node_safe_to_kill(self, node_ip: str) -> bool:
...
     def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
-        pool_label_selector = self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
-        pool_nodes = self._core_api.list_node().items
-
-        return {
-            get_node_ip(node): node
-            for node in pool_nodes
-            if not self.pool or node.metadata.labels.get(pool_label_selector, None) == self.pool
-        }
+        # TODO(https://jira.yelpcorp.com/browse/CLUSTERMAN-659)
```
nemacysts: nit: this comment should look something like:

```suggestion
        # TODO(CLUSTERMAN-659): ${SHORT_SUMMARY}
```

where `${SHORT_SUMMARY}` is just a couple words about what actually needs to be done

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -248,6 +247,30 @@ def _get_pods_info(
...
+        unschedulable_pods: List[KubernetesPod] = []
+        excluded_pods_by_ip: Mapping[str, List[KubernetesPod]] = defaultdict(list)
+
+        exclude_daemonset_pods = self.pool_config.read_bool(
+            "exclude_daemonset_pods",
+            default=staticconf.read_bool("exclude_daemonset_pods", default=False),
+        )
+        label_selector = f"{self.pool_label_key}={self.pool}"
+
+        for pod in self._core_api.list_pod_for_all_namespaces(label_selector=label_selector).items:
```
nemacysts, line 263: its a little unfortunate that this is the only line that's changing, yet we're duplicating the rest of the function

how long do we anticipate needing this duplication? if its going to be a while, we should maybe do something like having a single version of this function take a handle to a partial function that we've already given the label selector args to, having a single version of this function take/use **kwargs for the list of args we give list_pod_for_all_namespaces, or having  having a single version of this function take in the list of pods it will work on (and this way, we avoid having duplicated code hanging around for a long time)

clusterman/kubernetes/kubernetes_cluster_connector.py
akshaysharma096: Noted @nemacysts I will fix this.

clusterman/kubernetes/kubernetes_cluster_connector.py
akshaysharma096: It won't be a while for us to change the definition back to the correct one @nemacysts I will try to make it more better tho!

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -215,14 +215,13 @@ def _is_node_safe_to_kill(self, node_ip: str) -> bool:
...
     def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
-        pool_label_selector = self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
-        pool_nodes = self._core_api.list_node().items
-
-        return {
-            get_node_ip(node): node
-            for node in pool_nodes
-            if not self.pool or node.metadata.labels.get(pool_label_selector, None) == self.pool
-        }
+        # TODO(CLUSTERMAN-659): ${Switch to using just pool_label_key once the new node labels are applied everywhere}
```
nemacysts: nit:
```suggestion
        # TODO(CLUSTERMAN-659): Switch to using just pool_label_key once the new node labels are applied everywhere
```

nemacysts (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
akshaysharma096: Yep done 👍

nemacysts (approved):

piax93 (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -215,14 +215,13 @@ def _is_node_safe_to_kill(self, node_ip: str) -> bool:
...
-
-        return {
-            get_node_ip(node): node
-            for node in pool_nodes
-            if not self.pool or node.metadata.labels.get(pool_label_selector, None) == self.pool
-        }
+        # TODO(CLUSTERMAN-659): Switch to using just pool_label_key once the new node labels are applied everywhere
+        node_label_selector = self.pool_config.read_string(
+            "node_label_key", default=self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
+        )
```
piax93, line 221: nit, actually this doesn't really make things much different, as right now the logic is already "get all nodes, and then filter the ones with pool_label_key==self.pool", so I'd say there's no real added risk in just using `pool_label_key` directly.

piax93: ignore me, I should read tickets first.

ilkinmammadzada (approved):

PR #180, opened 2022-09-02
node_migration: configuration and event parsing

piax93:
### Description
Follow up to #179 with a bit more code to load and pass to places the migration settings to be support as part of pool setup.
Haven't added any of this to the documentation yet and I expect to do so once things are actually functional, instead of just having `# TODO: everything` all of the place.

### Testing Done
Just written unit tests, none of this still does anything interesting.

### DIFF
--- a/clusterman/batch/node_migration.py
+++ b/clusterman/batch/node_migration.py
@@ -42,7 +42,7 @@ from clusterman.util import setup_logging
 class NodeMigration(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin):
     notify_emails = ["compute-infra@yelp.com"]
 
-    POOL_SCHEDULER = "kube"
+    POOL_SCHEDULER = "kubernetes"
     POOL_SETTINGS_PARENT = "node_migration"
     EVENT_FETCH_BATCH = 10
 
--- a/clusterman/batch/node_migration.py
+++ b/clusterman/batch/node_migration.py
@@ -11,12 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import itertools
 from time import time
-from typing import Any
 from typing import Collection
+from typing import Generator
+from typing import Optional
 
 import colorlog
 import staticconf
+from botocore.exceptions import ClientError
 from yelp_batch.batch import batch_command_line_arguments
 from yelp_batch.batch import batch_configure
 from yelp_batch.batch_daemon import BatchDaemon
@@ -30,6 +33,8 @@ from clusterman.config import get_pool_config_path
 from clusterman.config import load_cluster_pool_config
 from clusterman.config import POOL_NAMESPACE
 from clusterman.config import setup_config
+from clusterman.migration.event import MigrationEvent
+from clusterman.migration.settings import WorkerSetup
 from clusterman.util import get_pool_name_list
 from clusterman.util import setup_logging
 
@@ -37,14 +42,15 @@ from clusterman.util import setup_logging
 class NodeMigration(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin):
     notify_emails = ["compute-infra@yelp.com"]
 
-    POOL_SCHEDULER = "kubernetes"
+    POOL_SCHEDULER = "kube"
     POOL_SETTINGS_PARENT = "node_migration"
+    EVENT_FETCH_BATCH = 10
 
     @batch_command_line_arguments
     def parse_args(self, parser):
         arg_group = parser.add_argument_group("NodeMigration batch options")
         add_env_config_path_arg(arg_group)
-        add_cluster_arg(arg_group)
+        add_cluster_arg(arg_group, required=True)
 
     @batch_configure
     def configure_initial(self):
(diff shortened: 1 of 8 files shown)


clusterman/batch/node_migration.py
```diff
@@ -30,21 +33,24 @@
...
+from clusterman.migration.settings import WorkerSetup
 from clusterman.util import get_pool_name_list
 from clusterman.util import setup_logging
 
 
 class NodeMigration(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin):
     notify_emails = ["compute-infra@yelp.com"]
 
-    POOL_SCHEDULER = "kubernetes"
+    POOL_SCHEDULER = "kube"
```
ilkinmammadzada: it should be kubernetes

piax93: I thought so, then testing got me confused cause it's just `kube` there, I'll get it fixed.

clusterman/batch/node_migration.py
```diff
@@ -53,7 +59,9 @@ def configure_initial(self):
         self.sqs_client = sqs()
         self.migration_workers = {}
         self.migration_configs = {}
+        self.events_in_progress = set()
         self.run_interval = staticconf.read_int("batches.node_migration.run_interval_seconds", 60)
+        self.event_visibilty_timeout = staticconf.read_int("batches.node_migration.event_visibilty_timeout", 15 * 60)
```
nemacysts: nit: its nice to have the unit of time in the name of the config option for clarity (e.g, just like the `run_interval_seconds` option above)

piax93: good catch, I'll fix this

clusterman/batch/node_migration.py
```diff
@@ -97,7 +151,7 @@ def run(self):
             if "event_queue" in config["trigger"]:
                 event_queues.add(config["trigger"]["event_queue"])
```
ilkinmammadzada: Currently we store all SQS URLs in common configuration file (clusterman.yaml)
Should we move this URL as well?

piax93: I thought it could come handy to have the option of setting it individually per-pool, but it's true that passing cluster and pool inside the events themselves kind of nullifies the utility of having the option. I'll rework things a little.

clusterman/batch/node_migration.py
```diff
@@ -53,36 +58,97 @@ def configure_initial(self):
         self.sqs_client = sqs()
         self.migration_workers = {}
         self.migration_configs = {}
+        self.events_in_progress = set()
+        self.pools_accepting_events = set()
+        self.event_queue = staticconf.read_string(f"clusters.{self.options.cluster}.migration_event_queue_url")
         self.run_interval = staticconf.read_int("batches.node_migration.run_interval_seconds", 60)
```
ilkinmammadzada, line 63: nit: We will use this `service` for choosing victim node. Probably we need to check again when new node come up. Long story short default value may be `average node creation time` which is 5 mins IIRC

piax93: node monitoring will be done by workers, the main loop here will just check that migration worker processes are healthy.

clusterman/migration/event.py
```diff
@@ -0,0 +1,54 @@
...
+
+class MigrationCondition(NamedTuple):
+    trait: ConditionTrait
+    target: Union[str, int, List[str]]
+
+
+class MigrationEvent(NamedTuple):
+    event_id: str
+    event_receipt: str
+    cluster: str
```
ilkinmammadzada, line 36: Why may we need clustername in the event?

clusterman/batch/node_migration.py
```diff
@@ -53,36 +58,97 @@ def configure_initial(self):
...
+            if pool in self.migration_configs:
+                return WorkerSetup.from_config(self.migration_configs[pool])
+        except Exception as e:
+            self.logger.exception(f"Bad migration configuration for pool {pool}: {e}")
+        return None
+
+    def fetch_event_queue(self) -> Collection[MigrationEvent]:
+        """Tail a collection of SQS queue for migration trigger events"""
+        events = map(MigrationEvent.from_event, self._fetch_all_from_queue())
+        return set(events) - self.events_in_progress
```
ilkinmammadzada, line 113: it never set. Probably you will have in the next commits. just in case

clusterman/migration/event.py
piax93: Makes it easier to make sure that events are meant to be received by a certain batch instance, in case we mess up something on the client side, or decide to share the same queue for multiple clusters.

clusterman/batch/node_migration.py
piax93: yep, I'll get `events_in_progress` populated once I get to work on the worker-process management code

clusterman/migration/event.py
ilkinmammadzada: We currently have a separate queue for all clusters. It would probably be better to keep them as is. By the way CLIs are also different for cluster. I mean we know about cluster. when send/receive event

ilkinmammadzada: `node migration` instance will also run on specific cluster. IMHO For now it would be better for each cluster to have its own `migration` so that they are all standard.

piax93: true that we'll have things separated by cluster in all places, but at the same time it doesn't hurt to have this information replicated in the events, it's just extra insurance.

ilkinmammadzada (approved):

PR #181, opened 2022-09-02
Kubernetes draining feature

ilkinmammadzada:
### Description

We have draining feature for Mesos, but we didn't implement it for Kubernetes.
Basically, it clean pods on victim node before terminating on AWS side.
Draining respects PDBs with threshold. Action (after threshold) can be configured.

### Testing Done

New test cases were added and some of existing were changed.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -79,13 +79,13 @@ class DrainingClient:
         )
 
     def submit_instance_for_draining(
-            self,
-            instance: InstanceMetadata,
-            sender: Type[AWSResourceGroup],
-            scheduler: str,
-            pool: str,
-            agent_id: str,
-            draining_start_time: arrow.Arrow,
+        self,
+        instance: InstanceMetadata,
+        sender: Type[AWSResourceGroup],
+        scheduler: str,
+        pool: str,
+        agent_id: str,
+        draining_start_time: arrow.Arrow,
     ) -> None:
         return self.client.send_message(
             QueueUrl=self.drain_queue_url,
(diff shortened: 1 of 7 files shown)


clusterman/draining/queue.py
```diff
@@ -276,7 +297,28 @@ def process_drain_queue(
                     self.submit_host_for_termination(host_to_process)
             elif host_to_process.scheduler == "kubernetes":
                 logger.info(f"Kubernetes host to drain and submit for termination: {host_to_process}")
-                self.submit_host_for_termination(host_to_process, delay=0)
+                spent_time = arrow.now() - host_to_process.draining_start_time
+                pool_config = staticconf.NamespaceReaders(
+                    POOL_NAMESPACE.format(pool=host_to_process.pool, scheduler="kubernetes")
+                )
+                force_terminate = pool_config.read_bool("draining.force_terminate", False)
+                draining_time_threshold_seconds = pool_config.read_int("draining.draining_time_threshold_seconds", 1800)
```
nemacysts, line 306: do we need these options for handling spot interruptions? 

we should be draining as fast as we can since aws will essentially do that for us anyway once they shutdown the box to reclaim it anyway

ilkinmammadzada: Actually this sections works only for scaling-down reason. We don't do any actions for spot interruptions (for k8s).

nemacysts: ah, i see - the main usage for draining in mesos was for handling spot interruptions (hence my confusion)

are we planning on using this code for handling spot interruption as well or would i still need to finish out https://github.com/Yelp/clusterman/pull/157 (or setting up the aws node termination handler)?

ilkinmammadzada: I see. I will work on spot interruptions right after churning and draining features.
I agree with you about `should be draining as fast as`. So I think it will be better delete (not evict) pods for spot warning (Or trying to evict with short period which doesn't make sense too much)

piax93 (approved):
LGTM

piax93 (approved):

PR #186, opened 2022-09-07
pool_manager: consider uptime in selecting killable nodes

piax93:
### Description
* Adds a new setting which can be used to indicate a grace period for new hosts being bootstrapped into a pool before they are prioritized for termination. This is off by default so that can selectively be applied to pools which may need it to prevent capacity flapping.
* In doing so I also changed the kill priority around staleness, as I while reading this I thought that it'd make more sense to kill node which are voluntarily marked as stale rather than ones which are just healthy but idling. Happy to revert if this reasoning is wrong.

### Testing Done
Just fixed/extended unit testing. Of course will do slow roll out if approved.

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -49,6 +49,7 @@ from clusterman.util import read_int_or_inf
 
 AWS_RUNNING_STATES = ("running",)
 MIN_CAPACITY_PER_GROUP = 1
+MAX_MIN_NODE_SCALEIN_UPTIME_SECONDS = 15 * 60  # 15 minutes
 SFX_RESOURCE_GROUP_MODIFICATION_FAILED_NAME = "clusterman.resource_group_modification_failed"
 logger = colorlog.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


docs/source/configuration.rst
```diff
@@ -156,7 +157,9 @@ The following is an example configuration file for a particular Clusterman pool:
 The ``resource-groups`` section provides information for loading resource groups in the pool manager.
 
 The ``scaling_limits`` section provides global pool-level limits on scaling that the autoscaler and
-other Clusterman commands should respect.
+other Clusterman commands should respect. The field ``min_node_scalein_uptime_seconds`` is an optional
+setting allowing to indicate a timespan in which freshly bootstrapped nodes are deprioritized in the
+selection for termination.
```
nemacysts, line 162: huh, we didn't previously document any of the scaling_limits params? :rip:

clusterman/autoscaler/pool_manager.py
```diff
@@ -546,12 +551,13 @@ def _prioritize_killable_nodes(self, killable_nodes: List[ClusterNodeMetadata])
 
         def sort_key(
             node_metadata: ClusterNodeMetadata,
-        ) -> Tuple[int, int, int, int, int, int]:
+        ) -> Tuple[int, int, int, int, int, int, int]:
             return (
                 0 if node_metadata.agent.is_frozen else 1,
                 0 if node_metadata.agent.state == AgentState.ORPHANED else 1,
-                0 if node_metadata.agent.state == AgentState.IDLE else 1,
                 0 if node_metadata.instance.is_stale else 1,
```
nemacysts, line 558: i think we'll be able to get rid of this once we've gotten rid of our SFRs (which we only use for mesos stuff) - after that, I don't think any of the staleness code will be used since, afaict, it's only used when we're rolling spot fleets

(only mentioning this since the pr description mentions updating the priority for stale nodes)

nemacysts (approved):

clusterman/autoscaler/pool_manager.py
piax93: Gotcha. The main reason of the priority change was that I wanted the new sort component to sit in front of `AgentState.IDLE`, but `stale` was in the middle, and in my eyes something manually marked as "bad" should have definitely have higher priority than uptime or task count.

PR #187, opened 2022-09-07
Reload kubernetes client on each try

ilkinmammadzada:
### Description

We couldn't access k8s api since client loaded in reload_state(which is too expensive).
We moved client loading to new method and used is for draining.

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -438,6 +438,8 @@ def process_queues(cluster_name: str) -> None:
 
     logger.info("Polling SQS for messages every 5s")
     while True:
+        if kube_operator_client:
+            kube_operator_client.reload_client()
         draining_client.clean_processing_hosts_cache()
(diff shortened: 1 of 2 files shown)


clusterman/draining/queue.py
```diff
@@ -438,6 +438,7 @@ def process_queues(cluster_name: str) -> None:
 
     logger.info("Polling SQS for messages every 5s")
     while True:
+        kube_operator_client.reload_client()
```
piax93: This may be set to `None` in case line 428 throws an error. Also I feel like that since this loop is much fast than the autoscaler, refreshing the client ever 5s (+ queue processing time) is a bit much.

ilkinmammadzada: Yeah, it will run every 5s, but I checked performance of `reload_client`. it just loads configuration file (~2 ms)

piax93 (approved):

PR #188, opened 2022-09-07
Build clusterman cli debian package for jammy

nemacysts:
As the tin says: this enables us to build the clusterman debian package for jammy hosts


debian/control
```diff
@@ -4,6 +4,6 @@ Build-Depends:
     dh-virtualenv,
 
 Package: clusterman
-Depends: python3.6, ${misc:Depends}, ${python:Depends}, ${shlibs:Depends}
+Depends: python3.7, ${misc:Depends}, ${python:Depends}, ${shlibs:Depends}
```
nemacysts: we depended on py36, but we tested py37

debian/rules
```diff
@@ -5,9 +5,24 @@ ifneq ($(shell echo ${CI}),true)
 	YELP_DH_INSTALL = install -d completions/usr debian/package/.
 endif
 
+# Use the dpkg version comparison algorithm (which is easily in reach) to compare the build
+# system version to the version from which python3-distutils is required (18.04 and higher),
+# and conditionally populate a substitution variable which is referenced in the control file.
+# On bionic, where Ubuntu's python3.8 is used, this is a virtual package for python3-distutils
```
nemacysts:
```suggestion
# On bionic, where Ubuntu's python3.7 is used, this is a virtual package for python3-distutils
```

package/itest/ubuntu.sh
```diff
@@ -29,7 +29,8 @@ export TZ=US/Pacific
 ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 apt-get update && apt-get install -y software-properties-common
-apt-get install -y --force-yes python3.7 python3-pip python3-yaml awscli
+add-apt-repository ppa:deadsnakes/ppa
+apt-get install -y --force-yes python3.7 python3-pip python3-yaml awscli python3.7-distutils
```
nemacysts, line 33: not really sure we i had to add distutils here - afaik, this should have been installed as part of my debian/{rules,control} change

(the rest of these should also probably be in debian/control too so that they'd be automatically installed, probably)

tox.ini
```diff
@@ -54,6 +54,9 @@ commands =
     sphinx-build -b html -d build/doctrees source build/html
 
 [testenv:virtualenv_run-dev]
+deps =
+    -rrequirements.txt
+    -rrequirements-dev.txt
```
nemacysts, line 59: this is unrelated, but it was annoying not having things like pre-commit available in the dev venv

package/itest/ubuntu.sh
```diff
@@ -29,7 +29,8 @@ export TZ=US/Pacific
 ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 apt-get update && apt-get install -y software-properties-common
-apt-get install -y --force-yes python3.7 python3-pip python3-yaml awscli
+add-apt-repository ppa:deadsnakes/ppa
```
nemacysts: !8ball this should be gated in an `if !YELP` conditional?

package/itest/ubuntu.sh
```diff
@@ -29,7 +29,8 @@ export TZ=US/Pacific
...
+apt-get install -y --force-yes python3.7 python3-pip python3-yaml awscli python3.7-distutils libatlas3-base
```
nemacysts: not really sure we i had to add distutils here - afaik, this should have been installed as part of my debian/{rules,control} change

(the rest of these should also probably be in debian/control too so that they'd be automatically installed, probably)

tzhu-yelp: agree

tzhu-yelp (approved):

debian/rules
```diff
@@ -5,9 +5,24 @@ ifneq ($(shell echo ${CI}),true)
 	YELP_DH_INSTALL = install -d completions/usr debian/package/.
 endif
 
+# Use the dpkg version comparison algorithm (which is easily in reach) to compare the build
+# system version to the version from which python3-distutils is required (18.04 and higher),
+# and conditionally populate a substitution variable which is referenced in the control file.
+# On bionic, where Ubuntu's python3.7 is used, this is a virtual package for python3-distutils
+# (in bionic-updates). On jammy, this is a real package providing distutils for python3.8 from deadsnakes.
```
jvperrin: `python3.8`? Isn't it 3.7 instead?

package/itest/ubuntu.sh
jvperrin: Hmmm, this seems fine to me to just use anyways, and I imagine quite a few other builds do the same

package/itest/ubuntu.sh
jvperrin: Agreed, yeah putting these as proper package dependencies would be much better if they are all truly needed, although things like `awscli` vs `aws-cli` might make that a bit more annoying

jvperrin (approved):
Overall looks great, thanks for uh... suffering through adding this, I really do appreciate it! 😬

debian/rules
nemacysts: err, i copy-paasta'd this from an internal repo and missed a py38->py37 reference :p

package/itest/ubuntu.sh
nemacysts: yea, imma try to mess around with this tomorrow - i was too tired of messing with this today to do something less disgusting :p

debian/control
BeryJu: I did notice this a while ago too that we seem to be expecting 3.6 as base in some places and 3.7 in some others

package/itest/ubuntu.sh
BeryJu: Stumbled over the same issue when I doing hacheck stuff, the paasta CLI in that container didnt work cause of missing distutils even though that should be installed via paasta's debian/control, probably fine to solve it the same way here

BeryJu (approved):

package/itest/ubuntu.sh
nemacysts: @jvperrin huh - should we be using `aws-cli` internally to grab our requirements patch? we've only ever referenced `awscli` in clusterman, so we've just been using the stock package

debian/rules
nemacysts:
```suggestion
# (in bionic-updates). On jammy, this is a real package providing distutils for python3.7 from deadsnakes.
```

debian/control
nemacysts: we really should just upgrade all of our stuff to the latest feasible version of python (which i think is 3.8 until we get rid of bionic?) - but i didn't wanna deal with that here :p

package/itest/ubuntu.sh
jvperrin: Oh ok then that's totally fine, and we're trying to migrate that way internally too so that aligns well!

debian/control
jvperrin: Totally makes sense, I think going with 3.7 here is reasonable to separate it out from the 3.7 -> 3.8 upgrade in a separate PR

package/itest/ubuntu.sh
nemacysts: not sure why we were installing awscli actually since we don't seem to actually use the cli in clusterman - i'm gonna go ahead and remove that

i think i also figured out why i had to re-add distutils here - i never referenced the `bionicOrLater:Depends` in debian/control

nemacysts: err, my eyes are bad - we do use the aws cli

nemacysts: nice, i think i got this working the right way now.

jvperrin (approved):

PR #182, opened 2022-09-05
cli tooling to trigger node migration events

piax93:
### Description
This adds a simple CLI interface allowing to trigger migration event, compatibly with what it started getting implemented in #180 and #191. I left the entry point commented out so that if a new version gets released with this in it, it won't confuse people.

### Testing Done
Just the usual unit testing.

PR #191, opened 2022-09-12
migration: switch to handling events with CRD

piax93:
### Description
Apologies for the big diff, wasn't easy to do in smaller chunks.
This replaces all the SQS stuff in the node migration batch to use a k8s custom resource instead. I also brought in some bits from #182 since I updated the condition logic quite a bit and wanted to have the input validation parts in place.

### Testing Done
* Added unit tests as always;
* Tried out the custom resource api calls in kubestage (no comment on the readability of the python k8s library...).

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import copy
 from collections import defaultdict
+from typing import Collection
 from typing import List
 from typing import Mapping
 from typing import Optional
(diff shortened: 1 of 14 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -184,6 +204,44 @@ def uncordon_node(self, node_name: str) -> bool:
...
+        :status MigrationStatus status: status to be set
+        """
+        assert self._migration_crd_api, "CRD client was not initialized"
+        try:
+            self._migration_crd_api.patch_cluster_custom_object(
+                name=event_name,
+                body={
+                    "metadata": {
+                        "labels": {
+                            MIGRATION_CRD_STATUS_LABEL: status.value,
```
ilkinmammadzada, line 239: I think it will be better keep labels immutable. If we want to patch/edit some fields, we should use spec.{field}.
What do you think store status in spec instead of label?

ilkinmammadzada (approved):
Small question about moving status from label to spec. otherwise LGTM

clusterman/kubernetes/kubernetes_cluster_connector.py
piax93: Keeping the status in labels allows querying the data much more easily. This was discussed with robj as well on the tech-spec review.

PR #194, opened 2022-09-13
Fix awscli conflict

nemacysts:
We install a forked aws-cli internally, and that conflicts with the upstream awscli.

`dpkg -I` on the bionic package says:
```
Depends: python3.7, libatlas3-base, libc6 (>= 2.26), libexpat1 (>= 2.1~beta3),
libgcc1 (>= 1:3.0), libstdc++6 (>= 5.2), libyaml-0-2, zlib1g (>= 1:1.2.0),
aws-cli
```

and on the jammy package:
```
Depends: python3.7, libatlas3-base, libc6 (>= 2.32), libgcc-s1 (>= 3.0),
libstdc++6 (>= 11), libyaml-0-2, awscli
```

I also installed this on the a kubestage box using `sudo apt-get install ./clusterman_4.10.4_amd64.deb` and got:
```
The following packages will be upgraded:
  clusterman
```
as opposed to what I get from installing the current mispackaged version
with `sudo apt-get install clusterman=4.10.4`
```
The following additional packages will be installed:
  awscli
The following NEW packages will be installed:
  awscli
The following packages will be upgraded:
  clusterman
```

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -16,6 +16,15 @@ else
     extra_substvars = -VbionicOrLater:Depends=""
 endif
 
+# and then do the same thing for awscli - of which we used to have a patched version called aws-cli pre-jammy.
+# once jammy boxes are the oldest things we install clusterman on, we can get rid of this and just include this
+# directly in debian/control
+ifeq ($(shell (. /etc/os-release && dpkg --compare-versions $$VERSION_ID "ge" "22.04" && echo yes || echo no)),yes)
+    extra_substvars = -VjammyOrLater:Depends="awscli"
+else
+    extra_substvars = -VjammyOrLater:Depends="aws-cli"
+endif
+
 
 %:
 	dh $@ --with python-virtualenv
--- a/debian/rules
+++ b/debian/rules
@@ -22,6 +22,10 @@ endif
 ifeq ($(shell (. /etc/os-release && dpkg --compare-versions $$VERSION_ID "ge" "22.04" && echo yes || echo no)),yes)
     extra_substvars = -VjammyOrLater:Depends="awscli"
 else
+	# aws-cli only exists internally, so lets make sure that we only use it internally
+    ifneq ($(shell echo ${PAASTA_ENV}), YELP)
+        extra_substvars = -VjammyOrLater:Depends="awscli"
+    endif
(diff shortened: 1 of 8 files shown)


jvperrin (approved):

jfongatyelp (approved):

BeryJu (approved):

nemacysts:
hmm, why is this suddenly failing to find pyyaml?

nemacysts:
sigh, lemme figure out why this is still upset

debian/rules
```diff
@@ -16,6 +16,19 @@ else
...
+# once jammy boxes are the oldest things we install clusterman on, we can get rid of this and just include this
+# directly in debian/control
+ifeq ($(shell (. /etc/os-release && dpkg --compare-versions $$VERSION_ID "ge" "22.04" && echo yes || echo no)),yes)
+    extra_substvars = -VjammyOrLater:Depends="awscli"
+else
+	# aws-cli only exists internally, so lets make sure that we only use it internally
+    ifneq ($(shell echo ${PAASTA_ENV}), YELP)
+        extra_substvars = -VjammyOrLater:Depends="awscli"
+    endif
+    extra_substvars = -VjammyOrLater:Depends="aws-cli"
```
jvperrin: Does this mean that `aws-cli` is used regardless and `awscli` is only used when outside Yelp? I guess I'm a bit confused why it's written this way around instead of swapping the two (and using an `ifeq` instead of `ifneq`), since it looks like this way around that `aws-cli` is always used, even outside Yelp.

nemacysts: err, i dun' goofed here - meant to add an else

🤦🏻

nemacysts: (i'll also swap the conditional)

BeryJu: Also do we actually *need* our forked version of awscli?

nemacysts: we don't - i mostly just don't want to deal with updating the puppet on all our pre-jammy boxes to use a non-forked version (since that's a ton of boxes)

BeryJu: oh right, this is for the clusterman-cli not for the part that runs in a container

jvperrin (approved):

BeryJu (approved):
:shipit:

PR #195, opened 2022-09-16
Implementing clean-up for warned kubernetes instances.

ilkinmammadzada:
### Description

We're currently doing nothing with the advance warning we get from AWS
for Spot instances in our K8s clusters.

### Testing

Test cases were added.

### DIFF
--- a/clusterman/draining/kubernetes.py
+++ b/clusterman/draining/kubernetes.py
@@ -7,43 +7,30 @@ from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesCluster
 log = colorlog.getLogger(__name__)
 
 
-def drain(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
-    """Cordons and safely evicts all tasks from a given node.
-    :param agent_id: a single node name to drain (as would be passed to kubectl drain)
+def drain(connector: Optional[KubernetesClusterConnector], node_name: str, disable_eviction: bool) -> bool:
+    """Cordons and evicts/deletes all tasks from a given node.
+    :param node_name: a single node name to drain (as would be passed to kubectl drain)
     :param connector: a kubernetes connector to connect kubernetes API
+    :param disable_eviction: Force drain to use delete (ignoring PDBs)
     :returns: bool
     """
     if connector:
-        log.info(f"Preparing to drain {agent_id}...")
-        return connector.drain_node(agent_id)
+        log.info(f"Preparing to drain {node_name}...")
+        return connector.drain_node(node_name, disable_eviction)
     else:
-        log.info(f"Unable to drain {agent_id} (no Kubernetes connector configured)")
+        log.info(f"Unable to drain {node_name} (no Kubernetes connector configured)")
         return False
 
 
-def uncordon(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
+def uncordon(connector: Optional[KubernetesClusterConnector], node_name: str) -> bool:
     """Cordons and safely evicts all tasks from a given node.
-    :param agent_id: a single node name to uncordon (as would be passed to kubectl uncordon)
+    :param node_name: a single node name to uncordon (as would be passed to kubectl uncordon)
     :param connector: a kubernetes connector to connect kubernetes API
     :returns: bool
     """
     if connector:
-        log.info(f"Preparing to uncordon {agent_id}...")
-        return connector.uncordon_node(agent_id)
+        log.info(f"Preparing to uncordon {node_name}...")
+        return connector.uncordon_node(node_name)
     else:
-        log.info(f"Unable to uncordon {agent_id} (no Kubernetes connector configured)")
-        return False
-
-
-def clean_node(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
-    """Cordons and forcibly delete all tasks from a given node.
-    :param agent_id: a single node name to clean
-    :param connector: a kubernetes connector to connect kubernetes API
-    :returns: bool
-    """
-    if connector:
-        log.info(f"Preparing to clean  on {agent_id}...")
-        return connector.clean_node(agent_id)
-    else:
-        log.info(f"Unable to clean {agent_id} (no Kubernetes connector configured)")
+        log.info(f"Unable to uncordon {node_name} (no Kubernetes connector configured)")
         return False
--- a/clusterman/draining/kubernetes.py
+++ b/clusterman/draining/kubernetes.py
@@ -33,3 +33,17 @@ def uncordon(connector: Optional[KubernetesClusterConnector], agent_id: str) -> 
     else:
         log.info(f"Unable to uncordon {agent_id} (no Kubernetes connector configured)")
         return False
+
+
+def clean_node(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
+    """Cordons and forcibly delete all tasks from a given node.
+    :param agent_id: a single node name to clean
+    :param connector: a kubernetes connector to connect kubernetes API
+    :returns: bool
+    """
+    if connector:
+        log.info(f"Preparing to clean  on {agent_id}...")
+        return connector.clean_node(agent_id)
+    else:
+        log.info(f"Unable to clean {agent_id} (no Kubernetes connector configured)")
+        return False
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -356,7 +356,7 @@ class DrainingClient:
                         should_resend_to_queue = True
                 else:
                     # case 3
-                    if k8s_drain(kube_operator_client, host_to_process.agent_id, disable_eviction=disable_eviction):
+                    if k8s_drain(kube_operator_client, host_to_process.agent_id, disable_eviction):
                         self.submit_host_for_termination(host_to_process, delay=0)
                         should_add_to_cache = True
                     else:  # case 4
@@ -390,7 +390,7 @@ class DrainingClient:
         for host in hosts_to_remove:
             del self.draining_host_ttl_cache[host]
 
-    def process_warning_queue(self, kube_operator_client: Optional[KubernetesClusterConnector]) -> None:
+    def process_warning_queue(self) -> None:
         host_to_process = self.get_warned_host()
         if host_to_process:
             logger.info(f"Processing spot warning for {host_to_process.hostname}")
@@ -508,7 +508,7 @@ def process_queues(cluster_name: str) -> None:
         if kube_operator_client:
             kube_operator_client.reload_client()
         draining_client.clean_processing_hosts_cache()
-        draining_client.process_warning_queue(kube_operator_client=kube_operator_client)
+        draining_client.process_warning_queue()
         draining_client.process_drain_queue(
             mesos_operator_client=mesos_operator_client,
             kube_operator_client=kube_operator_client,
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -77,7 +77,7 @@ class Host(NamedTuple):
     receipt_handle: str
     agent_id: str = ""
     pool: str = ""
-    termination_reason: TerminationReason = TerminationReason.SCALING_DOWN
+    termination_reason: str = TerminationReason.SCALING_DOWN.value
     draining_start_time: str = arrow.now().for_json()
     scheduler: str = "mesos"
 
@@ -124,6 +124,7 @@ class DrainingClient:
                     "instance_id": instance.instance_id,
                     "ip": instance.ip_address,
                     "pool": pool,
+                    "termination_reason": TerminationReason.SCALING_DOWN.value,
                     "scheduler": scheduler,
                 }
             ),
@@ -149,6 +150,7 @@ class DrainingClient:
                     "ip": host.ip,
                     "pool": host.pool,
                     "scheduler": host.scheduler,
+                    "termination_reason": host.termination_reason,
                 }
             ),
         )
@@ -179,6 +181,7 @@ class DrainingClient:
                     "ip": host.ip,
                     "pool": host.pool,
                     "scheduler": host.scheduler,
+                    "termination_reason": host.termination_reason,
                 }
             ),
         )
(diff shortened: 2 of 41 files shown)


clusterman/draining/kubernetes.py
```diff
@@ -33,3 +33,17 @@ def uncordon(connector: Optional[KubernetesClusterConnector], agent_id: str) ->
     else:
         log.info(f"Unable to uncordon {agent_id} (no Kubernetes connector configured)")
         return False
+
+
+def clean_node(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
+    """Cordons and forcibly delete all tasks from a given node.
```
piax93: Just to clarify, this means it doesn't care about PDBs, right?

piax93 (approved):

clusterman/draining/kubernetes.py
ilkinmammadzada: yeap, definitely! We use clean_node when we received spot termination warning form AWS. We should remove all pods asap. Because AWS terminates nodes in 2 minutes.

clusterman/draining/kubernetes.py
```diff
@@ -33,3 +33,17 @@ def uncordon(connector: Optional[KubernetesClusterConnector], agent_id: str) ->
     else:
         log.info(f"Unable to uncordon {agent_id} (no Kubernetes connector configured)")
         return False
+
+
+def clean_node(connector: Optional[KubernetesClusterConnector], agent_id: str) -> bool:
```
nemacysts: as i mentioned in the `freeze_agent` PR, we should probably rename this to something more aligned with k8s terminology imo (e.g., `force_drain_node` or something) - "cleaning" a node isn't exactly standard k8s language and it's something that we'll need to explain to new yelpers in the future when they're looking at this code for whatever reason (or when we're talking to them about this code) :)

clusterman/draining/queue.py
```diff
@@ -415,34 +434,44 @@ def host_from_instance_id(
...
         return None
     try:
         ip = instance_data[0]["PrivateIpAddress"]
     except KeyError:
         logger.warning(f"No primary IP found for {instance_id}")
         return None
+    try:
+        agent_id = instance_data[0]["PrivateDnsName"]
+    except KeyError:
+        logger.warning(f"No Dns name found for {instance_id}")
```
nemacysts:
```suggestion
        logger.warning(f"No DNS name found for {instance_id} - not draining.")
```

clusterman/draining/queue.py
```diff
@@ -373,29 +376,45 @@ def clean_processing_hosts_cache(self) -> None:
...
 
             # we should definitely ignore termination warnings that aren't from this
             # cluster or maybe not even paasta instances...
             if host_to_process.group_id in spot_fleet_resource_groups:
-                logger.info(f"Sending spot warned host to drain: {host_to_process.hostname}")
+                logger.info(f"Sending spot fleet warned host to drain: {host_to_process.hostname}")
                 self.submit_host_for_draining(host_to_process)
+            if host_to_process.group_id in autoscaling_resource_groups:
+                logger.info(f"Cleaning-up and sending spot warned host to terminate: {host_to_process.hostname}")
+                k8s_clean_node(kube_operator_client, host_to_process.agent_id)
```
nemacysts: hmm, should we do something like what we were doing for mesos and just have this place things to drain in a drain queue rather than doing the draining in-line here?

clusterman/draining/queue.py
```diff
@@ -373,29 +376,45 @@ def clean_processing_hosts_cache(self) -> None:
...
                     )
 
             # we should definitely ignore termination warnings that aren't from this
             # cluster or maybe not even paasta instances...
             if host_to_process.group_id in spot_fleet_resource_groups:
-                logger.info(f"Sending spot warned host to drain: {host_to_process.hostname}")
+                logger.info(f"Sending spot fleet warned host to drain: {host_to_process.hostname}")
                 self.submit_host_for_draining(host_to_process)
+            if host_to_process.group_id in autoscaling_resource_groups:
+                logger.info(f"Cleaning-up and sending spot warned host to terminate: {host_to_process.hostname}")
```
nemacysts: same thing as above: it'd be preferable to use standard terminology over `clean`

clusterman/draining/queue.py
```diff
@@ -373,29 +376,45 @@ def clean_processing_hosts_cache(self) -> None:
...
+                k8s_clean_node(kube_operator_client, host_to_process.agent_id)
+                self.submit_host_for_termination(host_to_process, 0)
```
nemacysts: suggestion:
```suggestion
                self.submit_host_for_termination(host_to_process, delay=0)
```
makes this line clearer at a glance

clusterman/draining/queue.py
```diff
@@ -415,34 +434,44 @@ def host_from_instance_id(
         logger.warning(f"No instance data found for {instance_id}")
         return None
     try:
-        sfr_ids = [tag["Value"] for tag in instance_data[0]["Tags"] if tag["Key"] == "aws:ec2spot:fleet-request-id"]
+        group_ids = [
+            tag["Value"]
+            for tag in instance_data[0]["Tags"]
+            if tag["Key"] in {"aws:ec2spot:fleet-request-id", "aws:ec2:fleet-id", }
```
nemacysts: nit: would be good to extract `{"aws:ec2spot:fleet-request-id", "aws:ec2:fleet-id", }` to a named constant

clusterman/draining/queue.py
```diff
@@ -415,34 +434,44 @@ def host_from_instance_id(
...
                 break
     except KeyError as e:
-        logger.warning(f"SFR tag key not found: {e}")
-        sfr_ids = []
-    if not sfr_ids:
-        logger.warning(f"No SFR ID found for {instance_id}")
+        logger.warning(f"SFR/SF tag key not found: {e}")
+        group_ids = []
+    if not group_ids:
+        logger.warning(f"No SFR/SF ID found for {instance_id}")
```
nemacysts: thoughts on rewording this a tiny bit:
```suggestion
        logger.exception(f"Spot tag key not found - is this Spot Fleet/ASG correctly configured?")
        group_ids = []
    if not group_ids:
        logger.warning(f"Not draining {instance_id}: no Spot ID found - is this actually a Spot instance?")
```

might clear things up for new folks looking at these logs

clusterman/draining/queue.py
```diff
@@ -415,34 +434,44 @@ def host_from_instance_id(
...
         return None
     try:
         ip = instance_data[0]["PrivateIpAddress"]
     except KeyError:
         logger.warning(f"No primary IP found for {instance_id}")
         return None
+    try:
+        agent_id = instance_data[0]["PrivateDnsName"]
```
nemacysts, line 468: just double-checking: all ec2 instances will have this right - including the SFR boxes we use for mesos?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -129,29 +129,46 @@ def get_unschedulable_pods(
...
 
-    def drain_node(self, agent_id: str) -> bool:
+    def clean_node(self, node_name: str) -> bool:
         try:
-            logger.info(f"Cordoning {agent_id}...")
-            if not self.cordon_node(agent_id):
+            logger.info(f"Cordoning {node_name}...")
+            self.cordon_node(node_name)
+        except Exception as e:
+            logger.warning(f"Failed to cordon {node_name} continuing to delete pods anyway: {e}")
```
nemacysts:
```suggestion
            logger.exception(f"Failed to cordon {node_name} - continuing to delete pods anyway.")
```

clusterman/draining/kubernetes.py
nemacysts: given our new focus on PDBs, might be good to include this as a parenthetical

e.g., `Cordons and forcibly (i.e., ignoring PDBs) delete all tasks from a given node.`

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -129,29 +129,46 @@ def get_unschedulable_pods(
...
+            return False
+
+    def drain_node(self, node_name: str) -> bool:
+        try:
+            logger.info(f"Cordoning {node_name}...")
+            if not self.cordon_node(node_name):
+                return False
+            logger.info(f"Evicting pods on {node_name}...")
+            if not self._evict_tasks_from_node(node_name):
                 logger.info("Some pods couldn't be evicted")
```
nemacysts: on mobile right now so I can't just do a GH suggestion - but thoughts on having this include the node name as well?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -129,29 +129,46 @@ def get_unschedulable_pods(
...
+            logger.warning(f"Failed to clean {node_name}: {e}")
+            return False
+
+    def drain_node(self, node_name: str) -> bool:
+        try:
+            logger.info(f"Cordoning {node_name}...")
+            if not self.cordon_node(node_name):
+                return False
+            logger.info(f"Evicting pods on {node_name}...")
+            if not self._evict_tasks_from_node(node_name):
```
nemacysts: what happens if this return False? will we retry later or just wait for AWS to kill everything for us?

clusterman/draining/kubernetes.py
ilkinmammadzada: I agree with you. I will try to use same method (drain) with new parameter (disable-eviction [bool]).

clusterman/draining/queue.py
ilkinmammadzada: Yeah, I listed EC2s in AWS UI, sort (asc, desc) by Private DNS Name, I saw that all them have it. Is it enough?

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Sorry, I didn't understand this one. this is already `node_name`

clusterman/draining/kubernetes.py
ilkinmammadzada: @nemacysts I created another PR to refactor kubernetes drain: https://github.com/Yelp/clusterman/pull/200

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: I ignored result of cordon.

ilkinmammadzada:
@nemacysts Thanks for all your valuable comments! I think I fixed all of them.

clusterman/draining/queue.py
ilkinmammadzada: I changed logic as requested. It looks better now. Thanks

clusterman/draining/queue.py
ilkinmammadzada: PrivateDnsName - agent_id is optional now. We will send these nodes (without PrivateDnsName) to termination directly.

nemacysts: yea - I was mostly just curious as to when this would not be set, but the AWS docs say `This element remains empty until the instance enters a running state` so this is fine :)

clusterman/draining/queue.py
```diff
@@ -377,27 +393,42 @@ def process_warning_queue(self) -> None:
...
-                logger.info(f"Sending spot warned host to drain: {host_to_process.hostname}")
+            if (
+                host_to_process.group_id in spot_fleet_resource_groups
+                or host_to_process.group_id in autoscaling_resource_groups
+            ):
+                logger.info(f"Sending warned host to drain: {host_to_process.hostname}")
                 self.submit_host_for_draining(host_to_process)
             else:
-                logger.info(f"Ignoring spot warned host because not in our SFRs: {host_to_process.hostname}")
+                logger.info(f"Ignoring warned host because not in our target: {host_to_process.hostname}")
```
nemacysts: nit:
```suggestion
                logger.info(f"Ignoring warned host because not in our target group: {host_to_process.hostname}")
```

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -270,14 +270,14 @@ def create_node_migration_resource(
...
-        logger.info(f"{len(pods)} pods being evicted")
+        logger.info(f"{len(pods)} pods being evicted/deleted")
         for pod in pods:
             try:
                 if disable_eviction:
                     self._delete_pod(pod)
                 else:
                     self._evict_pod(pod)
-                logger.info(f"{pod.metadata.name} ({pod.metadata.namespace}) was deleted")
+                logger.info(f"{pod.metadata.name} ({pod.metadata.namespace}) was evicted/deleted")
```
nemacysts: suggestion: would be nice to have this include the node name - but that might be annoying to pull out so no worries :)

clusterman/draining/queue.py
```diff
@@ -316,6 +329,9 @@ def process_drain_queue(
                     default=DEFAULT_DRAINING_TIME_THRESHOLD_SECONDS,
                 )
                 should_resend_to_queue = False
+                disable_eviction = (
+                    True if host_to_process.termination_reason == TerminationReason.SPOT_INTERRUPTION else False
+                )
```
nemacysts: nit:
```suggestion
                disable_eviction = host_to_process.termination_reason == TerminationReason.SPOT_INTERRUPTION
```

we'll already get a `bool` from the comparison, so we can skip the `True if X else False` dance here :)

clusterman/draining/queue.py
```diff
@@ -54,6 +57,15 @@
...
 DEFAULT_DRAINING_TIME_THRESHOLD_SECONDS = 1800
+EC2_FLEET_KEYS = {
+    "aws:ec2spot:fleet-request-id",
+    "aws:ec2:fleet-id",
+}
+
+
+class TerminationReason(enum.Enum):
+    SCALING_DOWN = "scaling down"
+    SPOT_INTERRUPTION = "spot interruption"
```
nemacysts, line 68: suggestion: if we don't need the string values anywhere, we can always use https://docs.python.org/3/library/enum.html#using-automatic-values here instead

nemacysts (approved):

clusterman/draining/queue.py
ilkinmammadzada: OMG! how did i do this?! :D

clusterman/draining/queue.py
ilkinmammadzada: I noticed that I didn't add Reason information to message in the queue.

PR #197, opened 2022-09-21
node_migration: uptime monitoring worker

piax93:
### Description
This is most of the juice of the actual migration logic, with worker processes being spawned and submitting nodes for draining. This just contains the uptime monitoring bit since event processing will also require some more juggling the PoolManager to support filtering nodes with label selectors.
I'm also not very satisfied of the naive `is_capacity_satisfied` approach to establish pool health, as I feel it's a bit too simplistic. Very open to hear some feedback on that.
Regarding process handling, that's very naive as well at the moment. I ticketed myself about refining it a little in a future iteration.

### Testing Done
None apart from writing unit tests.

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -192,25 +192,33 @@ class PoolManager:
 
         if not dry_run:
             if self.draining_enabled:
-                assert self.draining_client  # make mypy happy
-                for group_id, node_metadatas in marked_nodes_by_group.items():
+                for node_metadatas in marked_nodes_by_group.values():
                     for node_metadata in node_metadatas:
-                        self.draining_client.submit_instance_for_draining(
-                            node_metadata.instance,
-                            sender=cast(
-                                Type[AWSResourceGroup],
-                                self.resource_groups[group_id].__class__,
-                            ),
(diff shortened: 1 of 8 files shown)


clusterman/autoscaler/pool_manager.py
```diff
@@ -192,26 +192,34 @@ def prune_excess_fulfilled_capacity(
...
                         [node_metadata.instance.instance_id for node_metadata in node_metadatas]
                     )
 
+    def submit_for_draining(self, node_metadata: ClusterNodeMetadata) -> None:
+        """Submit collection of nodes for draining
+
+        :param ClusterNodeMetadata node_metadata: node to be drained
+        """
+        assert self.draining_client  # make mypy happy
+        group_id = node_metadata.instance.group_id
```
piax93, line 210: As far as I know, this should always be set, so getting it from `marked_nodes_by_group` was superfluous. Otherwise I can just make it an optional parameter to the method to maintain as much of the existing logic as possible.

ilkinmammadzada: Yeah, I think so. Let's keep eyes on logs in deployment time, just in case.

ilkinmammadzada (approved):

PR #198, opened 2022-09-23
node migration: event migration worker

piax93:
### Description
Last few pieces of the puzzle after #197. Basically a very similar implementation, just with filtering done based on the information contained in the migration event.

To make things look a bit nicer I had to modify a couple of bits in other place, so will need a bit of care in deploying this one (broke changes into different commits to make then a bit easier to review):
* updated k8s cluster manager class to actually use label selectors rather than list all hosts and then filter by label later.
* I needed some good way to enforce timeout, so I reworked the SIGALRM based implementation used by CLI handlers to be a bit more generic
* Same goes for the autoscaler toggles in dynamodb, so that I could reuse the code in the worker function.

This will very likely need a bit more polish on top of this, but should be a decent base implementation for the system.

### Testing Done
As always, just unit tests for now.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -274,6 +274,15 @@ class KubernetesClusterConnector(ClusterConnector):
         except Exception as e:
             logger.error(f"Failed creating migration event resource: {e}")
 
+    def has_enough_capacity_for_pods(self) -> bool:
+        """Checks whether there are unschedulable pods due to insufficient resources
+
+        :return: True if no unschedulable pods are due to resource constraints
+        """
+        return not any(
+            reason == PodUnschedulableReason.InsufficientResources for _, reason in self.get_unschedulable_pods()
+        )
+
     def _evict_tasks_from_node(self, hostname: str) -> bool:
         all_evicted = True
         pods_to_evict = [
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -69,6 +69,7 @@ class KubernetesClusterConnector(ClusterConnector):
     _unschedulable_pods: List[KubernetesPod]
     _excluded_pods_by_ip: Mapping[str, List[KubernetesPod]]
     _pods_by_ip: Mapping[str, List[KubernetesPod]]
+    _label_selectors: Collection[str]
 
     def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) -> None:
         super().__init__(cluster, pool)
@@ -79,6 +80,10 @@ class KubernetesClusterConnector(ClusterConnector):
         )
         self._nodes_by_ip = {}
         self._init_crd_client = init_crd
+        self._label_selectors = []
+        if self.pool:
+            pool_label_selector = self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
+            self._label_selectors.append(f"{pool_label_selector}={self.pool}")
 
     def reload_state(self) -> None:
         logger.info("Reloading nodes")
@@ -106,6 +111,16 @@ class KubernetesClusterConnector(ClusterConnector):
             )
             if self._init_crd_client
             else None
+        )
+
+    def set_label_selectors(self, label_selectors: Collection[str], add_to_existing: bool = False) -> None:
+        """Set label selectors for node listing purposes
+
+        :param Collection[str] label_selectors: list of selectors (joined with logic and)
+        :param bool add_to_existing: if set add to existing selectors rather than replacing
+        """
+        self._label_selectors = sorted(
+            (set(self._label_selectors) | set(label_selectors)) if add_to_existing else set(label_selectors)
         )
 
     def get_num_removed_nodes_before_last_reload(self) -> int:
@@ -377,14 +392,9 @@ class KubernetesClusterConnector(ClusterConnector):
         return True
 
     def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
-        pool_label_selector = self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
-        pool_nodes = self._core_api.list_node().items
-
-        return {
-            get_node_ip(node): node
-            for node in pool_nodes
-            if not self.pool or node.metadata.labels.get(pool_label_selector, None) == self.pool
-        }
+        kwargs = {"label_selector": ",".join(self._label_selectors)} if self._label_selectors else {}
+        pool_nodes = self._core_api.list_node(**kwargs).items
+        return {get_node_ip(node): node for node in pool_nodes}
 
     def _get_pods_info(
         self,
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -73,7 +73,7 @@ class KubernetesClusterConnector(ClusterConnector):
     _unschedulable_pods: List[KubernetesPod]
     _excluded_pods_by_ip: Mapping[str, List[KubernetesPod]]
     _pods_by_ip: Mapping[str, List[KubernetesPod]]
-    _label_selectors: Collection[str]
+    _label_selectors: List[str]
 
     def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) -> None:
         super().__init__(cluster, pool)
@@ -117,10 +117,10 @@ class KubernetesClusterConnector(ClusterConnector):
             else None
         )
 
-    def set_label_selectors(self, label_selectors: Collection[str], add_to_existing: bool = False) -> None:
+    def set_label_selectors(self, label_selectors: List[str], add_to_existing: bool = False) -> None:
         """Set label selectors for node listing purposes
 
-        :param Collection[str] label_selectors: list of selectors (joined with logic and)
+        :param List[str] label_selectors: list of selectors (joined with logic and)
         :param bool add_to_existing: if set add to existing selectors rather than replacing
         """
         self._label_selectors = sorted(
--- a/clusterman/migration/worker.py
+++ b/clusterman/migration/worker.py
@@ -27,7 +27,6 @@ from clusterman.autoscaler.toggle import disable_autoscaling
 from clusterman.autoscaler.toggle import enable_autoscaling
 from clusterman.interfaces.types import ClusterNodeMetadata
 from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
-from clusterman.kubernetes.util import PodUnschedulableReason
 from clusterman.migration.event import MigrationEvent
 from clusterman.migration.settings import WorkerSetup
 from clusterman.util import limit_function_runtime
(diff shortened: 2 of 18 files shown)


clusterman/migration/worker.py
```diff
@@ -74,7 +82,11 @@ def _monitor_pool_health(manager: PoolManager, timeout: float, drained: Collecti
         draining_happened = draining_happened or not any(
             node.agent.agent_id == connector.get_agent_metadata(node.instance.ip_address).agent_id for node in drained
         )
-        if draining_happened and manager.is_capacity_satisfied() and not connector.get_unschedulable_pods():
+        if (
+            draining_happened
+            and manager.is_capacity_satisfied()
+            and not (check_pods and connector.get_unschedulable_pods())
```
ilkinmammadzada: Looks little bit risky to me. Our migration will stop if there is any pod can't be scheduled. What do you think consider "PodUnschedulableReason" as well in your conditions?

piax93: Good call, I'll look in what info that can give.

clusterman/migration/worker.py
```diff
@@ -54,12 +61,19 @@ def __getattr__(self, attr):
...
+
+def _monitor_pool_health(
+    manager: PoolManager, timeout: float, drained: Collection[ClusterNodeMetadata], check_pods: bool = True
+) -> bool:
     """Monitor pool health after nodes were submitted for draining
 
     :param PoolManager manager: pool manager instance
     :param float timeout: timestamp after which giving up
     :param Collection[ClusterNodeMetadata] drained: nodes which were submitted for draining
+    :param bool check_pods: check that pods can successfully be schedules
```
nemacysts: nit:
```suggestion
    :param bool check_pods: check that pods can successfully be scheduled
```

clusterman/migration/worker.py
```diff
@@ -69,7 +83,17 @@ def _monitor_pool_health(manager: PoolManager, timeout: float, drained: Collecti
...
+        if (
+            draining_happened
+            and manager.is_capacity_satisfied()
+            and not (
+                check_pods
+                and any(
+                    reason == PodUnschedulableReason.InsufficientResources
+                    for _, reason in connector.get_unschedulable_pods()
+                )
+            )
```
nemacysts: thoughts on extracting this to a variable (or function if we wanna keep the short-circuiting behavior)? 10 lines for an if-statement is getting pretty long (+ the multiple levels of indentation) and naming the `any()` check will help this be more self-documenting

clusterman/migration/worker.py
```diff
@@ -131,4 +157,44 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
...
     :param WorkerSetup worker_setup: migration setup
     """
-    pass  # TODO
+    prescaling_applied = False
+    manager = PoolManager(migration_event.cluster, migration_event.pool, SUPPORTED_POOL_SCHEDULER, fetch_state=False)
+    connector = cast(KubernetesClusterConnector, manager.cluster_connector)
+    connector.set_label_selectors(migration_event.label_selectors, add_to_existing=True)
+    manager.reload_state()
+    try:
+        if worker_setup.disable_autoscaling:
```
nemacysts, line 158: i'm still a little worried about us pausing the autoscaler for migrations (even with pre-scaling) - if someone triggers a migration before peak traffic times when we're scaled down, i can easily see us overwhelming the pre-scaled capacity

clusterman/migration/worker.py
```diff
@@ -131,4 +157,44 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
...
+        logger.error(f"Issue while processing migration event {migration_event}: {e}")
+        raise
+    finally:
+        if worker_setup.disable_autoscaling:
+            logger.info(f"Re-enabling autoscaling for {migration_event.cluster}:{migration_event.pool}")
+            enable_autoscaling(migration_event.cluster, migration_event.pool, SUPPORTED_POOL_SCHEDULER)
+        if worker_setup.prescaling and prescaling_applied and prescaled_capacity >= manager.target_capacity:
+            # Only restoring original capacity of the target remained equal or lower than
+            # our set target (i.e. the autoscaler didn't have to scale up in the meantime)
+            manager.modify_target_capacity(original_capacity)
```
nemacysts: hmm, i wonder if it's worth leaving this out: presumably clusterman will run relatively soon after autoscaling is re-enabled so at most we'll spend an extra 5 minutes (i *think* that's the longest period we run?) scaled up?

clusterman/util.py
```diff
@@ -418,3 +424,25 @@ def strtobool(val: str) -> int:
         return 0
     else:
         raise ValueError("invalid truth value %r" % (val,))
+
+
+def limit_function_runtime(
+    func: Callable[[], Any], max_runtime: Union[int, float], callback: Optional[Callable[[], None]] = None
```
nemacysts, line 430: super-nit: it might be nice to use [ParamSpec/TypeVar](https://docs.python.org/3/library/typing.html#typing.ParamSpec) to type `func` (actually, maybe we just need TypeVar since we only expect 0-arg functions to be passed)

clusterman/migration/worker.py
nemacysts: (not sure if this is just a temporary thing until we get to the autoscaler interference ideas mentioned in the tech spec for this)

clusterman/migration/worker.py
```diff
@@ -131,4 +157,44 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
...
+            logger.info(f"Disabling autoscaling for {migration_event.cluster}:{migration_event.pool}")
+            disable_autoscaling(
+                migration_event.cluster,
+                migration_event.pool,
+                SUPPORTED_POOL_SCHEDULER,
+                time.time() + worker_setup.expected_duration,
+            )
+        if worker_setup.prescaling:
+            nodes = manager.get_node_metadatas(AWS_RUNNING_STATES)
+            offset = worker_setup.prescaling.of(len(nodes))
```
nemacysts, line 168: just curious: how does this work? i don't see an `of` method in [WorkerSetup](https://github.com/Yelp/clusterman/blob/4fedeb246c182310ae2598335b46b7f11754109f/clusterman/migration/settings.py#L46:7)

clusterman/migration/worker.py
```diff
@@ -131,4 +157,44 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
...
+            avg_weight = mean(node.instance.weight for node in nodes)
+            original_capacity = manager.target_capacity
+            prescaled_capacity = round(manager.target_capacity + offset * avg_weight)
```
nemacysts: feel free to drop: should we parenthesize the math here or just trust that people will read this according to the precedence rules?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -377,14 +392,9 @@ def _is_node_safe_to_kill(self, node_ip: str) -> bool:
...
-        pool_nodes = self._core_api.list_node().items
-
-        return {
-            get_node_ip(node): node
-            for node in pool_nodes
-            if not self.pool or node.metadata.labels.get(pool_label_selector, None) == self.pool
-        }
+        kwargs = {"label_selector": ",".join(self._label_selectors)} if self._label_selectors else {}
+        pool_nodes = self._core_api.list_node(**kwargs).items
+        return {get_node_ip(node): node for node in pool_nodes}
```
nemacysts: nice!

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -108,6 +113,16 @@ def reload_client(self) -> None:
             else None
         )
 
+    def set_label_selectors(self, label_selectors: Collection[str], add_to_existing: bool = False) -> None:
+        """Set label selectors for node listing purposes
+
+        :param Collection[str] label_selectors: list of selectors (joined with logic and)
+        :param bool add_to_existing: if set add to existing selectors rather than replacing
+        """
+        self._label_selectors = sorted(
```
nemacysts, line 131: nit: does this need to be sorted? seems like we could `s/sorted/list` and save some cpu cycles :)

nemacysts (commented):
just left a couple comments/suggestions/nits - nothing blocking tho

clusterman/migration/worker.py
piax93: This is off by default. It will just have some use in very small dev pool, the ones with 3/4 hosts in which pre-scaling by 1 causes the autoscaler to go real mad.

clusterman/migration/worker.py
piax93: I was thinking this as well. Can't tell how sensible the autoscaler will be in certain cases, but I guess won't leave it running overscaled for days 🤞

clusterman/util.py
piax93: I don't really see a use for that though, since the parameter type is not re-referenced in any other place.

clusterman/migration/worker.py
piax93: The `prescaling` property is an instance of [`PoolPortion`](https://github.com/Yelp/clusterman/blob/master/clusterman/migration/settings.py#L46). Just an abstanction to transparently support both percentages and absolute values in the configuration file.

(yes, sometimes I definitely go for a bit too much OOP - and now the whole Java community will be angry with me for dynamically setting a method instead of using a factory :P)

clusterman/migration/worker.py
piax93: Will add some parenthesis, never trust math.

clusterman/kubernetes/kubernetes_cluster_connector.py
piax93: Thanks! I've see this was already in the works for #177, but I guess that's not getting shipped soon 😢

clusterman/kubernetes/kubernetes_cluster_connector.py
piax93: It doesn't really need to be, but since it's a bit which can be useful to be human readable, for example in k8s audit logs, it'll be easier if the labels are not shuffled every time due to `set`. And it's going to be used very sporadically, with surely less than 4 strings in the set, so negligible performance impact.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -83,6 +84,10 @@ def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) ->
         )
         self._nodes_by_ip = {}
         self._init_crd_client = init_crd
+        self._label_selectors = []
+        if self.pool:
+            pool_label_selector = self.pool_config.read_string("pool_label_key", default="clusterman.com/pool")
```
ilkinmammadzada: Currently I see paasta.yelp.com/pool and yelp.com/pool=default labels on node. @akshaysharma096 can confirm that which one will be kept.

piax93: As far as I know, all poll configuration set this value. I think the currently used contention is `yelp.com/pool`.

piax93: But I'd opt to keep `clusterman.com/pool` here as a default value, since this is an open source project, and doesn't really need to be Yelp-branded.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -269,6 +284,15 @@ def create_node_migration_resource(
         except Exception as e:
             logger.error(f"Failed creating migration event resource: {e}")
 
+    def has_enough_capacity_for_pods(self) -> bool:
+        """Checks whether there are unschedulable pods due to insufficient resources
+
+        :return: True if no unschedulable pods are due to resource constraints
+        """
+        return not any(
+            reason == PodUnschedulableReason.InsufficientResources for _, reason in self.get_unschedulable_pods()
```
ilkinmammadzada, line 298: Cool

ilkinmammadzada (approved):

PR #200, opened 2022-09-25
Refactored kubernetes drain

ilkinmammadzada:
### Description

Implemented  disable_eviction flag for kubernetes drain. This enables force delete pods (ignoring PDBs) on node while draining. We will use force drain for spot interruption. 

### Testing Done

tests were changed as well

### DIFF
--- a/clusterman/draining/kubernetes.py
+++ b/clusterman/draining/kubernetes.py
@@ -7,29 +7,30 @@ from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesCluster
 log = colorlog.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


piax93 (commented):
LGTM. I just don't understand why renaming `agent_id` to `node_name`? (I mean, technically more correct in k8s terms, but `agent_id` is used in other places in this codebase).

ilkinmammadzada:
> LGTM. I just don't understand why renaming `agent_id` to `node_name`? (I mean, technically more correct in k8s terms, but `agent_id` is used in other places in this codebase).

@piax93 general components (who works with mesos and kubernetes) know it as "agent_id". But it is "node_name" in kubernetes language. We will refactor most terms after getting mesos rid off.

piax93 (approved):

PR #204, opened 2022-09-27
Avoided error while ec2 termination

ilkinmammadzada:
### Description

Currently clusterman crashed if any error (for ex: ClientError can be occurred) occurs while ec2 termination. 

### Solution

Handle exception silently. Autoscaler will submit it again in case of necessity.


piax93 (approved):
Let's go for this as an intermediate solution as per our offline discussion.

PR #205, opened 2022-09-28
avoid node migration overlaps for a pool

piax93:
This avoids 2 migration workers to be acting on the same pool at the same time, so that we don't risk causing any weird behaviour due to interference. It does so in a couple of ways:

* Processes are labelled based on `{type}:{cluster}:{pool}` (cluster is unnecessary given our deployment model, but makes logs easier to read), so that on any given pool at most 1 uptime monitoring process and 1 migration event process can be active.
* Worker methods are passed a lock, related to a specific pool, which they acquire before starting any mutation on the pool, so that uptime monitoring will pause if in the meantime an event gets triggered for the same pool.

I had to base this PR off of the branch of #198 cause I needed the worker function implementation to put the lock calls in place. Please prioritize the review of that one.

### DIFF
--- a/clusterman/batch/node_migration.py
+++ b/clusterman/batch/node_migration.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import time
+from collections import defaultdict
+from multiprocessing import Lock
 from typing import Callable
 from typing import Collection
 from typing import Dict
@@ -45,9 +47,6 @@ from clusterman.util import get_pool_name_list
 from clusterman.util import setup_logging
 
 
-WorkerProcessLabel = Union[str, MigrationEvent]
-
-
 class NodeMigration(BatchDaemon, BatchLoggingMixin, BatchRunningSentinelMixin):
     notify_emails = ["compute-infra@yelp.com"]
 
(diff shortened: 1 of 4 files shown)


ilkinmammadzada (approved):

clusterman/batch/node_migration.py
```diff
@@ -113,10 +136,12 @@ def _spawn_worker(self, label: WorkerProcessLabel, routine: Callable, *args, **k
             self.logger.warning(f"Worker labelled {label} already running, skipping")
             return False
         running_workers = sum(proc.is_alive() for proc in self.migration_workers.values())
-        if isinstance(label, MigrationEvent) and running_workers >= self.available_worker_slots:
+        if self._is_event_worker_label(label) and running_workers >= self.available_worker_slots:
             # uptime workers are prioritized skipping this check
             self.logger.warning(f"Too many worker processes running already ({running_workers}), skipping")
             return False
+        lock_key = label.split(self.WORKER_LABEL_SEPARATOR, 1)[1]
```
nemacysts, line 143: i'm personally a fan of avoiding explicit indexing, but that might just be a me thing and i guess
```python
        _, lock_key = label.split(self.WORKER_LABEL_SEPARATOR, maxsplit=1)
```
isn't really much different here

nemacysts (approved):

PR #206, opened 2022-09-28
show node migration info in "clusterman status" output

piax93:
* Add a pool label to the migration CRD resources so that they are easier to filter when we don't just want to list all the ones available in a cluster. Chosen the same label key being used for nodes (I know it's redundant info since the pool value is also contained in the event themselves, but I'd rather make k8s work a bit more than having to do client-side filtering here).
* Simply adds an extra section to the `clusterman status` output with the migration events currently being processed (if any). I put this behind its own parameter for the time being so that there's no risk of people encountering possible issues while doing other stuff. I guess it will eventually go with the `verbose` flag.

### DIFF
--- a/clusterman/cli/status.py
+++ b/clusterman/cli/status.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import argparse
 import sys
+from typing import Collection
 from typing import Iterable
 from typing import List
 from typing import Mapping
@@ -39,6 +40,9 @@ from clusterman.cli.util import timeout_wrapper
 from clusterman.interfaces.resource_group import ResourceGroup
 from clusterman.interfaces.types import AgentState
 from clusterman.interfaces.types import ClusterNodeMetadata
+from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
+from clusterman.migration.event import MigrationEvent
+from clusterman.migration.event_enums import MigrationStatus
 from clusterman.util import any_of
 from clusterman.util import autoscaling_is_paused
 from clusterman.util import ClustermanResources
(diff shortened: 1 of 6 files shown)


clusterman/cli/status.py
```diff
@@ -270,6 +295,13 @@ def add_status_parser(subparser, required_named_args, optional_named_args):  # p
         action="store_true",
         help="Only show information about orphaned instances (instances that are not in the Mesos cluster)",
     )
+    optional_named_args.add_argument(
+        "--migrations",
```
nemacysts: hmm, thoughts on `--show-migrations` just for a little bit of additional clarity?

clusterman/cli/status.py
```diff
@@ -76,6 +80,7 @@ class StatusJsonObject(TypedDict):
     target_capacity: float
     non_orphan_fulfilled_capacity: float
     resource_groups: List[ResourceGroupJsonObject]
+    migrations: Collection[MigrationEvent]
```
nemacysts: hmm, we should probably pick a style and be consistent - we've mixed using the typing ABCs with the concrete types all over the place (in paasta too, to be fair :p)

personally, I prefer having function params be typed using the abstract classes and everything else using the concrete ones - but i don't know if anyone else has opinions

that said, if we wanted to be consistent here: this should be 
```suggestion
    migrations: List[MigrationEvent]
```
which i think might necessitate changing the type in a couple other places too?

clusterman/cli/status.py
```diff
@@ -116,14 +121,26 @@ def _get_resource_groups_json(
...
-def _status_json(manager: PoolManager, get_node_metadatas: bool) -> StatusJsonObject:
+def _get_migrations(cluster: str, pool: str) -> Collection[MigrationEvent]:
+    """Get node migrations currently in flight for a pool
+
+    :param str cluster: cluster name
+    :param str pool: pool name:
+    :return: collection of migration events
+    """
+    connector = KubernetesClusterConnector(cluster, pool, init_crd=True)
+    return connector.list_node_migration_resources(statuses=[MigrationStatus.PENDING, MigrationStatus.INPROGRESS])
```
nemacysts: just talking out loud: would have been nice to make statuses a Set instead of a List so that we enforce uniqueness

nemacysts (approved):

clusterman/cli/status.py
piax93: The argument I'd have against this is pretty much the same I raised in #198 regarding sorting of label selectors, as using sets would end up causing the order of the selector values to randomly change at every call. Considering that at the moment this function is always invoked without user input, so it's pretty easy to verify that there are no duplicates, I'd leave it as it is.

PR #209, opened 2022-10-03
node_migration: add option to ignore pod health

piax93:
I know there are some efficiency improvements coming up with additional labelling of pods in our pools, but I think it's worth having an option like this to be used in our very large and general purpose pools, in which establishing the health based on pods is a bit of a stretch, and it's quite time/resource consuming to fetch that information out.


clusterman/autoscaler/pool_manager.py
```diff
@@ -84,12 +84,13 @@ def __init__(
         if fetch_state:
             self.reload_state()
 
-    def reload_state(self) -> None:
+    def reload_state(self, **cluster_connector_kwargs) -> None:
```
akshaysharma096, line 87: nice.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -94,20 +94,23 @@ def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) ->
...
         self._prev_nodes_by_ip = copy.deepcopy(self._nodes_by_ip)
         self._nodes_by_ip = self._get_nodes_by_ip()
-        logger.info("Reloading pods")
-        (self._pods_by_ip, self._unschedulable_pods, self._excluded_pods_by_ip,) = (
-            self._get_pods_info_with_label()
-            if self.pool_config.read_bool("use_labels_for_pods", default=False)
-            else self._get_pods_info()
-        )
+        if load_pods_info:
+            logger.info("Reloading pods")
```
akshaysharma096, line 106: I think the log statement should come before the if else clause?

akshaysharma096 (changes requested):

clusterman/kubernetes/kubernetes_cluster_connector.py
piax93: I don't think we should write in logs that it's reloading pods, when effectively it's not.

akshaysharma096: A better way is to add a verbose log of what strategy we are using to get the pods `self._pods_by_ip, self._unschedulable_pods, self._excluded_pods_by_ip` 

```py
if load_pods_info:
    logger.info("Reloading pods via label filtering")
else:
   logger.info("Reloading pods without any label filteribg")
```

piax93: But that's outside the scope of this change.

akshaysharma096: Cool!

akshaysharma096 (approved):
Ship!

PR #210, opened 2022-10-03
update mypy and a couple other quality of life improvements

piax93:
* Old version of mypy was being annoying regarding kwargs, the latest update is a bit better in that sense (and even more annoying on a lot of other bits, so don't know if I actually won my battle, but I've already put in the work so 🤷)
* We use py37 for the repo, so we can use the builtin `mock` module rather than importing the third party one (and this makes mypy much happier)
* While I was at it, there were a couple of warnings being raised while running unit tests which were trivial to fix, so I did that.


requirements-dev-minimal.txt
```diff
@@ -2,10 +2,14 @@ behave>=1.2.6
...
 pytest
 requirements-tools
-mock
-moto>=1.3.15
-mypy
+types-cachetools==5.2.1
+types-PyYAML==6.0.12
+types-requests==2.28.11
+types-retry==0.9.9
+types-urllib3==1.26.25
```
nemacysts: these should only be pinned in requirements-dev.txt
```suggestion
types-cachetools
types-PyYAML
types-requests
types-retry
types-urllib3
```

clusterman/migration/event.py
```diff
@@ -131,7 +133,7 @@ class MigrationEvent(NamedTuple):
 
     def __hash__(self) -> int:
         """Simplified object hash since resource_name should be unique"""
-        return self[:3].__hash__()
+        return cast(Tuple[str, str, str], self[:3]).__hash__()
```
nemacysts, line 136: huh, interesting that mypy couldn't figure this out

clusterman/aws/spot_fleet_resource_group.py
```diff
@@ -145,7 +145,7 @@ def _target_capacity(self) -> float:
         return self._configuration["SpotFleetRequestConfig"]["TargetCapacity"]
 
     @classmethod
-    def load(
+    def load(  # type: ignore
```
nemacysts: it'd be nice to add a comment to these type ignores in case someone wants to take the effort to remove them in the future (or in case the typing failures are linked to an existing mypy issue), but that might just be a luisp-strong-opinion

nemacysts (approved):

requirements-dev-minimal.txt
nemacysts: (this is the only blocking comment)

piax93: oops, nice catch, some copy pasta 😅

clusterman/aws/spot_fleet_resource_group.py
piax93: I'd love to, if only mypy would give me some better information about why it gets angry at this method signature. I tried all the permutations of removing the kwargs and turning the result type to a more generic type, but it still just complained with a "method signature doesn't match".

clusterman/migration/event.py
piax93: It was complaining with 

```
error: Invalid self argument "MigrationEvent" to attribute function "__hash__" with type "Callable[[MigrationEvent], int]"
```

It basically cannot figure out that `MigrationEvent` is effectively a tuple, and a slice of itself is no longer the same type.

PR #212, opened 2022-10-05
Removed log line what cause flood.

ilkinmammadzada:
### Description

We have many failed and succeeded pods in some pools (specially in default). This cause 2000+ lines logs. 

Another task for more proper logging related pods and nodes reloading: https://jira.yelpcorp.com/browse/CLUSTERMAN-697


piax93 (commented):
Shall we replace it with something like a counter? So we can just log something like "Excluding X pods from pool metadata".

ilkinmammadzada:
@piax93 yeap. I added new task to description. It's for now. We will add proper logging here in the next sprint.

piax93 (approved):

PR #216, opened 2022-10-07
Avoiding to mark orphan instances as killable

ilkinmammadzada:
### Description

Currently orphan instances are terminated by autoscaler with priority. 
Our average node bootstrapping time is 5mins. 
Autoscaler may terminate (before bootstrapping finishing) these instances in case scaling-down needed.

We have `terminate_expired_orphan_instances` what takes care of orphan instances. it has setting `orphan_instance_uptime_threshold_seconds` (default 30 mins) to define expired orphan instances.

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -553,6 +553,9 @@ class PoolManager:
             return False
         elif node_metadata.instance.is_cordoned:
             return False
+        # Orphan instances will be terminated by terminate_expired_orphan_instances
+        elif node_metadata.agent.state == AgentState.ORPHANED:
+            return False
         elif self.max_tasks_to_kill > node_metadata.agent.task_count:
             return True
         else:
@@ -563,10 +566,9 @@ class PoolManager:
 
         def sort_key(
             node_metadata: ClusterNodeMetadata,
-        ) -> Tuple[int, int, int, int, int, int, int]:
+        ) -> Tuple[int, int, int, int, int, int]:
             return (
                 0 if node_metadata.agent.is_frozen else 1,
-                0 if node_metadata.agent.state == AgentState.ORPHANED else 1,
                 0 if node_metadata.instance.is_stale else 1,
                 0 if node_metadata.instance.uptime.total_seconds() > self.min_node_scalein_uptime else 1,
                 0 if node_metadata.agent.state == AgentState.IDLE else 1,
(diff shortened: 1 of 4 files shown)


clusterman/autoscaler/pool_manager.py
```diff
@@ -80,6 +81,10 @@ def __init__(
             self.pool_config.read_int("scaling_limits.min_node_scalein_uptime_seconds", default=-1),
             MAX_MIN_NODE_SCALEIN_UPTIME_SECONDS,
         )
+        autoscaling_config = get_autoscaling_config(
+            POOL_NAMESPACE.format(pool=self.pool, scheduler=self.scheduler),
+        )
+        self.orphan_instance_uptime_threshold_seconds = autoscaling_config.orphan_instance_uptime_threshold_seconds
```
piax93: I don't know if re-using this parameter is a good idea, as this indicates the maximum time after which an orphan node should be considered unrepairable, while you are looking for a minimum grace period. Don't know if we may end up having any problem with it.

Also, we should probably rename `min_node_scalein_uptime_seconds ` to something like `min_node_scalein_uptime_priority_seconds`, since that goes to inform selection sorting rather than filtering (my bad for not being more explicit when I introduced it)

ilkinmammadzada: Actually I think again that we don't need this setting (orphan_instance_uptime_threshold_seconds) for marking instance non-killable. We can mark orphan non-killable since "terminate_expired_orphan_instances" will terminate them. What do you think?

I can change `min_node_scalein_uptime_seconds`

piax93: But then if a pool has `recycle_orphan_instances` set to False those node would end up living forever, wouldn't they?

ilkinmammadzada: Yeah, perfect catch!

ilkinmammadzada: > don't know if re-using this parameter is a good idea, as this indicates the maximum time after which an orphan node should be considered unrepairable, while you are looking for a minimum grace period. Don't know if we may end up having any problem with it.

Actually they are similar (maybe same). We just need threshold to mark orphan as killable after it reached.  And our recycling terminates these oprhans if it's enabled.

ilkinmammadzada: I thought again about orphan instances. We shouldn’t mark them as killable. Our cleaning logic will clean orphans. Yeap some pool owners can disable it, but there is no reason for it (who loves to keep orphan instances in pool?). I implemented flag for smooth deployment. Even I can remove flag. Additionally we have alert for orphan instances in case of necessity.

ilkinmammadzada:
I will try to catch issue in drainer side. Because marking orphan as killable makes sense. 
https://github.com/Yelp/clusterman/pull/220/files

PR #219, opened 2022-10-07
emit monitoring metrics for node-migration workers

piax93:
I guess this is a reasonable way to implement them, but very open to suggestions.

### DIFF
--- a/clusterman/migration/worker.py
+++ b/clusterman/migration/worker.py
@@ -132,7 +132,7 @@ def _drain_node_selection(
         for node in selection_chunk:
             logger.info(f"Recycling node {node.instance.instance_id}")
             manager.submit_for_draining(node)
-            node_uptime_gauge.set(node.instance.uptime.total_seconds(), {"ip": node.instance.ip_address})
+            node_uptime_gauge.set(node.instance.uptime.total_seconds())
             node_drain_counter.count()
         time.sleep(worker_setup.bootstrap_wait)
         if not _monitor_pool_health(
--- a/clusterman/migration/worker.py
+++ b/clusterman/migration/worker.py
@@ -30,6 +30,7 @@ from clusterman.interfaces.types import ClusterNodeMetadata
 from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
 from clusterman.migration.event import MigrationEvent
 from clusterman.migration.settings import WorkerSetup
+from clusterman.monitoring_lib import get_monitoring_client
 from clusterman.util import limit_function_runtime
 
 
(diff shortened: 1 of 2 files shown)


clusterman/migration/worker.py
```diff
@@ -112,14 +117,23 @@ def _drain_node_selection(
...
     chunk = worker_setup.rate.of(len(nodes))
     logger.info(f"{len(selected)} nodes of {manager.cluster}:{manager.pool} will be recycled")
+    job_timer.start()
     for i in range(0, len(selected), chunk):
         start_time = time.time()
         selection_chunk = selected[i : i + chunk]
         for node in selection_chunk:
             logger.info(f"Recycling node {node.instance.instance_id}")
             manager.submit_for_draining(node)
+            node_uptime_gauge.set(node.instance.uptime.total_seconds(), {"ip": node.instance.ip_address})
```
ilkinmammadzada: I think we should add more general label to create dashboard/statistics instead of ip. for ex: cluster/pool, OS version, and etc.

clusterman/migration/worker.py
```diff
@@ -112,14 +117,23 @@ def _drain_node_selection(
     """
     nodes = manager.get_node_metadatas(AWS_RUNNING_STATES)
     selected = sorted(filter(selector, nodes), key=worker_setup.precedence.sort_key)
+    if not selected:
+        return True
+    monitoring_info = {"cluster": manager.cluster, "pool": manager.pool}
+    node_drain_counter = get_monitoring_client().create_counter(SFX_NODE_DRAIN_COUNT, monitoring_info)
+    job_timer = get_monitoring_client().create_timer(SFX_MIGRATION_JOB_DURATION, monitoring_info)
+    node_uptime_gauge = get_monitoring_client().create_gauge(SFX_DRAINED_NODE_UPTIME, monitoring_info)
```
ilkinmammadzada, line 125: What do you think about creating metrics as properties of class (in the constructor) and use it inside method?
similar to autoscaler

clusterman/migration/worker.py
piax93: Node and pool will be included already by the base setup of the gauge. I thought having the node IP could be useful to backtrack anomalies, but if you don't think that's the cause I can just remove that extra info.

clusterman/migration/worker.py
piax93: There's no class is, the worker code is all functional.

clusterman/migration/worker.py
ilkinmammadzada: Got it, Yeah I think we should remove it. Because it may cause high cardinality in TSDB

clusterman/migration/worker.py
```diff
@@ -112,14 +117,23 @@ def _drain_node_selection(
     """
     nodes = manager.get_node_metadatas(AWS_RUNNING_STATES)
     selected = sorted(filter(selector, nodes), key=worker_setup.precedence.sort_key)
+    if not selected:
+        return True
+    monitoring_info = {"cluster": manager.cluster, "pool": manager.pool}
+    node_drain_counter = get_monitoring_client().create_counter(SFX_NODE_DRAIN_COUNT, monitoring_info)
+    job_timer = get_monitoring_client().create_timer(SFX_MIGRATION_JOB_DURATION, monitoring_info)
```
ilkinmammadzada, line 124: wow! library++

ilkinmammadzada (approved):

PR #220, opened 2022-10-10
reworked orphan instance handling

ilkinmammadzada:
### Description

Currently we forcibly terminate orphan (host without agent_id) instance in k8s draining. 

We should to be sure that instance is orphan before terminate it. Because delay can be occurred between autoscaler and drainer and instance can be non-orphan anymore. 
### Testing Done

Added unit tests

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -310,7 +310,7 @@ class DrainingClient:
         host_to_process = self.get_host_to_drain()
         if host_to_process and host_to_process.instance_id not in self.draining_host_ttl_cache:
             # We should add instance_id to cache only if we submit host for termination
-            should_add_to_cache = False
+            should_add_to_cache = True
             if host_to_process.scheduler == "mesos":
                 logger.info(f"Mesos host to drain and submit for termination: {host_to_process}")
                 try:
@@ -324,7 +324,6 @@ class DrainingClient:
                     logger.error(f"Failed to drain {host_to_process.hostname} continuing to terminate anyway: {e}")
                 finally:
                     self.submit_host_for_termination(host_to_process)
-                    should_add_to_cache = True
             elif host_to_process.scheduler == "kubernetes":
                 logger.info(f"Kubernetes host to drain and submit for termination: {host_to_process}")
                 spent_time = arrow.now() - arrow.get(host_to_process.draining_start_time)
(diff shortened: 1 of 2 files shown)


clusterman/draining/queue.py
```diff
@@ -336,36 +335,45 @@ def process_drain_queue(
...
                 )
+                #  flag to detect failed cordon for adding message again to queue
                 should_resend_to_queue = False
                 disable_eviction = host_to_process.termination_reason == TerminationReason.SPOT_INTERRUPTION.value
-
                 # Try to drain node; there are a few different possibilities:
-                #  0) Instance is orphan, it should be terminated
+                #  0) host is orphan, getting host information from AWS
+                #       a) host doesn't exist, don't need any action
+                #       b) host doesn't have agent_id (PrivateDnsName), submit it for termination
```
piax93, line 353: Just realized that `host_from_instance_id` load the `agent_id` from the instance metadata. I think that's not a solid way to do things, as instances will have a private DNS name regardless of the k8s agent running on them or not, so I suspect you'll never see it unset in here.

ilkinmammadzada: Yeah, right. It will continue with 0c case. 

CMAN will try to drain orphan node as normal node:

- Cordon - failed but will continue.
- Delete pods - node has 0 pod.
- Submit it for termination - EC2 will be removed.

ilkinmammadzada: There were two more ways:

- Check if node exist via cluster_connector (from in-memory): there may be race condition, which is currently also exist. And it's really problem for some pools.
- Check if node exist via cluster_connection (real time): little bit costly. 

Additionally I still need `host_from_instance_id ` to find node name for above actions.

piax93: Ok it will fall back to another case, but I don't think it's good to keep a false assumption around in code and comments.

ilkinmammadzada: tbh I don't understand where is exactly false assumption.
0) host is orphan, getting host information from AWS : If we receive host without agent_id, this really means host is/was orphan (when autoscaler submit it for draining).

piax93: Nevermind, I got confused between `host_to_process_fresh` and `host_to_process`

piax93: Actually no, `host_to_process_fresh.agent_id` is treated as it informs the k8s agent to be missing, but since it's loaded from AWS data (https://github.com/Yelp/clusterman/blob/2072e1cd44bf88716c57824410b121a5e72b9f01/clusterman/draining/queue.py#L489), that's not the case.

ilkinmammadzada: I don't think so. Because log line is "Host doesn't have agent_id". It would be "Host is orphan" if agent missing. 
But I'm open for suggestion. I'm ready to change log line or comment if you suggest something.

piax93: But the point is that, reading that information from AWS APIs, the `agent_id` is always going to be set. EC2 always gives a private DNS record for instances. So the implementation of `0b` is just unreachable code, and likely confusing if someone else will end up reading it.

ilkinmammadzada: Oh, got it. I didn't that all EC2s must have `PrivateDnsName`. So I'm gonna delete this line. Thanks!

ilkinmammadzada: @piax93 I checked before deleting line and I saw some EC2s without PrivateDnsName. They were in terminated and shutting-down state. I want to merge case 0a and 0b and ignore as 0a. What do you think? 
Because we may get the message late or duplicate.

piax93: Yes, as you pointed out, the only case in which the EC2 API will return incomplete data is if an instance is already terminated (or in the process of being terminated).

ilkinmammadzada: And in this case clusterman will behave weirdly:

- receiving orphan instance which is in "shutting-down" state (due to duplicate event)
- calling host_from_instance_id() and getting host but without agent_id
- sending host to drain

same process will continue until node deleted from AWS side.
But if EC2 stuck while termination message same process will continue indefinitely. 

that's why I wanted to keep 0b case as well.

piax93 (approved):

PR #222, opened 2022-10-10
Drainer message reading optimization

ilkinmammadzada:
Currently we have performance issue with drainer, therefore there are many messages in the queue specially in the following cases:

- time between peak and non-peak, because autoscaler submits many hosts to drain.
- spot market hot time, because drainer submits many hosts to drain (forcibly).

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -273,9 +273,11 @@ class DrainingClient:
         self,
         mesos_operator_client: Optional[Callable[..., Callable[[str], Callable[..., None]]]],
         kube_operator_client: Optional[KubernetesClusterConnector],
-    ) -> None:
+    ) -> bool:
+        message_exist = False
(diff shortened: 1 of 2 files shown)


piax93 (approved):

clusterman/draining/queue.py
```diff
@@ -505,6 +505,7 @@ def host_from_instance_id(
 def process_queues(cluster_name: str) -> None:
     draining_client = DrainingClient(cluster_name)
     cluster_manager_name = staticconf.read_string(f"clusters.{cluster_name}.cluster_manager")
+    essential_drain_delay = staticconf.read_bool(f"clusters.{cluster_name}.essential_drain_delay", False)
```
piax93: I would suggest calling it something like "always_delay_drain_processing", otherwise it may look like the field is expecting a timespan value.

piax93 (approved):

PR #224, opened 2022-10-12
added caching resource group names

ilkinmammadzada:
### Description

Resource group load operations is too costly (~3-6 seconds).
We should cache resource group name since we don't change them frequently

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -18,6 +18,8 @@ import socket
 import time
 from typing import Callable
 from typing import Dict
+from typing import Hashable
+from typing import MutableMapping
 from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
@@ -25,6 +27,7 @@ from typing import Set
 from typing import Type
 
 import arrow
+import cachetools
 import colorlog
 import staticconf
 from botocore.exceptions import ClientError
@@ -54,6 +57,7 @@ from clusterman.util import get_pool_name_list
 
 logger = colorlog.getLogger(__name__)
 DRAIN_CACHE_SECONDS = 1800
+DEFAULT_RESOURCE_GROUPS_CACHE_SECONDS = 0
 DEFAULT_FORCE_TERMINATION = False
 DEFAULT_GLOBAL_REDRAINING_DELAY_SECONDS = 15
 DEFAULT_DRAINING_TIME_THRESHOLD_SECONDS = 1800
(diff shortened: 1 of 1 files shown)


clusterman/draining/queue.py
```diff
@@ -91,9 +92,16 @@ def __init__(self, cluster_name: str) -> None:
...
+        self.resource_group_cache_seconds = staticconf.read_int(
+            f"clusters.{cluster_name}.resource_group_cache_seconds",
+            default=DEFAULT_RESOURCE_GROUPS_CACHE_SECONDS,
+        )
         self.drain_queue_url = staticconf.read_string(f"clusters.{cluster_name}.drain_queue_url")
         self.termination_queue_url = staticconf.read_string(f"clusters.{cluster_name}.termination_queue_url")
         self.draining_host_ttl_cache: Dict[str, arrow.Arrow] = {}
+        self.spot_fleet_resource_groups_cache: Set[str] = set()
+        self.autoscaling_resource_groups_cache: Set[str] = set()
+        self.resource_group_cache_expiration_time: arrow.Arrow = arrow.now()
```
piax93: This should be set at time 0, so you are sure to do a first pass.
That said, since cachetools is already installed for the repo, I'd honestly just move each set of IDs to it's own time-cached property with separate TTLs (as we can make it a bit longer for ASGs).

clusterman/draining/queue.py
```diff
@@ -422,33 +430,37 @@ def process_warning_queue(self) -> bool:
...
+                    pool_config = staticconf.NamespaceReaders(POOL_NAMESPACE.format(pool=pool, scheduler="mesos"))
+                    for resource_group_conf in pool_config.read_list("resource_groups"):
+                        spot_fleet_resource_groups |= set(
+                            SpotFleetResourceGroup.load(
+                                cluster=self.cluster,
+                                pool=pool,
+                                config=list(resource_group_conf.values())[0],
+                            ).keys()
+                        )
+                self.spot_fleet_resource_groups_cache = spot_fleet_resource_groups
```
piax93: Another optimization you could make is to pre-match the group_id you get for `host_to_process`, as I guess names for SFR will have a somewhat stable prefix (I think it's `sfr-` and some hex chars, but I'd have to verify), so if something "looks like" an SFR host, you just load SFR IDs, and if not you just load ASGs names.

ilkinmammadzada: Probably this code will be removed soon. Additionally, I think it's little bit dangerous because someone can change this logic (pre-match) without notify us

piax93:
Sorry if I'm insisting on this, but the implementation using cachetools can be way simpler and readable, for example for SFRs:

```diff
diff --git a/clusterman/draining/queue.py b/clusterman/draining/queue.py
index 89e5442..c300170 100644
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -98,6 +98,13 @@ class DrainingClient:
             f"clusters.{cluster_name}.warning_queue_url",
             default=None,
         )
+        self.sfr_resource_group_cache = cachetools.TTLCache(
+            maxsize=1,
+            ttl=staticconf.read_int(
+                f"clusters.{cluster_name}.sfr_resource_group_cache_seconds",
+                default=DEFAULT_RESOURCE_GROUPS_CACHE_SECONDS,
+            ),
+        )

     def submit_instance_for_draining(
         self,
@@ -412,6 +419,22 @@ class DrainingClient:
         for host in hosts_to_remove:
             del self.draining_host_ttl_cache[host]

+    @property
+    @cachetools.cachedmethod(lambda self: self.sfr_resource_group_cache)
+    def spot_fleet_resource_groups(self):
+        result = set()
+        for pool in get_pool_name_list(self.cluster, "mesos"):
+            pool_config = staticconf.NamespaceReaders(POOL_NAMESPACE.format(pool=pool, scheduler="mesos"))
+            for resource_group_conf in pool_config.read_list("resource_groups"):
+                result |= set(
+                    SpotFleetResourceGroup.load(
+                        cluster=self.cluster,
+                        pool=pool,
+                        config=list(resource_group_conf.values())[0],
+                    ).keys()
+                )
+        return result
+
     def process_warning_queue(self) -> bool:
         message_exist = False
         host_to_process = self.get_warned_host()
```

and then you can juts reference `self.spot_fleet_resource_groups` and let cachetools handle all the rest

ilkinmammadzada:
@piax93 Thanks for suggestion! I implemented as you requested.
But test-external is unhappy. Could you please help with it?
`Incompatible types in assignment (expression has type "TTLCache[<nothing>, <nothing>]", variable has type "Set[str]")`

piax93 (approved):

PR #228, opened 2022-10-17
U/ilkinmammadzada/clusterman 681 avoid duplications

ilkinmammadzada:
### Description

Currently we add instance_id to cache only if draining is success. 
This may cause having same instance_id multiple times. 
We are changing logic little bit to have instance_id in cache in any cases, but check if it's first try 
or re-try due to failed draining. 

### Testing Done

fixed unit tests

### DIFF
--- a/clusterman/draining/queue.py
+++ b/clusterman/draining/queue.py
@@ -82,6 +82,7 @@ class Host(NamedTuple):
     receipt_handle: str
     agent_id: str = ""
     pool: str = ""
+    attempt: int = 1
     termination_reason: str = TerminationReason.SCALING_DOWN.value
     draining_start_time: str = arrow.now().for_json()
     scheduler: str = "mesos"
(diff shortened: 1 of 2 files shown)


clusterman/draining/queue.py
```diff
@@ -388,34 +391,28 @@ def process_drain_queue(
...
                 else:  # case 3
-                    should_add_to_cache = self._drain_k8s_host(kube_operator_client, host_to_process, disable_eviction)
-                    should_resend_to_queue = not should_add_to_cache
-
-                if should_resend_to_queue:
-                    logger.info(
-                        f"Delaying re-draining {host_to_process.instance_id} for {redraining_delay_seconds} seconds"
-                    )
-                    self.submit_host_for_draining(host_to_process, redraining_delay_seconds)
+                    if not self._drain_k8s_host(kube_operator_client, host_to_process, disable_eviction):
```
piax93: This can become an `elif` and save 1 block of indentation.

ilkinmammadzada: @piax93 done.

piax93 (approved):

PR #229, opened 2022-10-17
added node_migration termination reason

ilkinmammadzada:
### Description

Currently we have two reasons for draining (scale-down and spot interruption)

We should add node_migration as well to get some stats. 

### Testing Done

fixed unit tests


piax93 (approved):
Code looks alright to me, just some tests need fixing

PR #230, opened 2022-10-17
removed unnecessary freeze action

ilkinmammadzada:
### Description

Currently we freeze (actually cordon) agent (node) in autoscaler before submitting for draining. But we don't need it anymore. Because we are cordoning nodes in drainer and it's not responsibility of autoscaler. Additionally we have some other places where we submit instance for draining (for ex: node_migration). We should cordon node in single place (drainer)

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -562,7 +562,7 @@ class PoolManager:
             node_metadata: ClusterNodeMetadata,
         ) -> Tuple[int, int, int, int, int, int, int]:
             return (
-                0 if node_metadata.agent.is_frozen else 1,
+                0 if node_metadata.agent.is_draining else 1,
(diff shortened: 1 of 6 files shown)


piax93:
I think this helps the autoscaler though, cause in case on the next iteration the drainer hasn't picked up the drain message yet, the node won't be taken into account for the total pool capacity (I think).

ilkinmammadzada:
Actually it takes into account frozen node as well. We added freeze logic before draining improvement. Reason was avoiding pod scheduling to node which will be terminated soon.

piax93 (approved):

ilkinmammadzada:
Let's keep it draft until deploy "clusterman specific cordon/uncordon". Because we are using "is_frozen" to prioritize killable nodes..

piax93 (approved):

piax93 (approved):

PR #232, opened 2022-10-17
made cordon/uncordon clusterman specific

ilkinmammadzada:
### Description

Currently we don't use any specific key to cordon node (in drainer).
But we should use specific key to be sure that node was cordoned by CMAN. 
And then we can use it (instead of is_frozen) to prioritize killable nodes. 

this PR will unblock https://github.com/Yelp/clusterman/pull/230

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -219,33 +219,37 @@ class KubernetesClusterConnector(ClusterConnector):
             return False
 
     def cordon_node(self, node_name: str) -> bool:
+        now = str(arrow.now().timestamp)
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": True,
-                    },
-                },
+            node = self._get_node_by_node_name(node_name)
+            if not node:
+                return False
+            taints = (
+                list(filter(lambda x: x.key != CLUSTERMAN_TERMINATION_TAINT_KEY, node.spec.taints))
+                if node.spec.taints
+                else []
             )
+            taints.append({"effect": "NoSchedule", "key": CLUSTERMAN_TERMINATION_TAINT_KEY, "value": now})
+            self._core_api.patch_node(name=node_name, body={"spec": {"taints": taints}})
             return True
         except ApiException as e:
-            logger.warning(f"Failed to cordon {node_name}: {e}")
+            logger.warning(f"Failed to cordon {node_name}: {e.status} - {e.reason}")
             return False
 
     def uncordon_node(self, node_name: str) -> bool:
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": False,
-                    },
-                },
+            node = self._get_node_by_node_name(node_name)
+            if not node:
+                return False
+            taints = (
+                list(filter(lambda x: x.key != CLUSTERMAN_TERMINATION_TAINT_KEY, node.spec.taints))
+                if node.spec.taints
+                else []
             )
+            self._core_api.patch_node(name=node_name, body={"spec": {"taints": taints}})
             return True
         except ApiException as e:
-            logger.warning(f"Failed to uncordon {node_name}: {e}")
+            logger.warning(f"Failed to uncordon {node_name}: {e.status} - {e.reason}")
             return False
 
     def list_node_migration_resources(self, statuses: List[MigrationStatus]) -> Set[MigrationEvent]:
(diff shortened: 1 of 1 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,30 +219,24 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
...
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": True,
-                    },
-                },
-            )
+            node = self._core_api.read_node(node_name)
```
piax93: Won't this information be available in `self._nodes_by_ip`? So you don't need to do an extra read.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,30 +219,24 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
             return False
 
     def cordon_node(self, node_name: str) -> bool:
+        now = str(arrow.now().timestamp)
+        key = CLUSTERMAN_TERMINATION_TAINT_KEY
```
piax93: nit, I think it's more readable if you reference the constant directly where it's used rather than renaming it locally.

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: I was thinking to get information real-time to minimise race condition, but yeah agree with you. I'm gonna change it and read real-time in case of necessity.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,33 +219,37 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
...
+        now = str(arrow.now().timestamp)
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": True,
-                    },
-                },
+            node = next(filter(lambda x: x.metadata.name == node_name, self._nodes_by_ip.values()))
```
piax93: This raises an exception if the node is not found, you need to add a default return to `next` to avoid that.

Also, can't we just turn the name into an IP by resolving it, or doing a string transformation, since we use internal DNS names from AWS, and those contain the IP.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,33 +219,37 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
...
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": False,
-                    },
-                },
+            node = next(filter(lambda x: x.metadata.name == node_name, self._nodes_by_ip.values()))
+            if not node:
+                return False
```
piax93: On top of the comment above, this should go into a `get_node_by_hostname` method to avoid duplicating code. Or even move the whole thing into a `get_node_taints` method.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,33 +219,39 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
...
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": False,
-                    },
-                },
+            ip = self._get_ip_from_node_name(node_name)
+            node = self._nodes_by_ip[ip] if ip in self._nodes_by_ip else None
```
piax93, line 243: This is the same as
```suggestion
            node = self._nodes_by_ip.get(ip)
```

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -219,33 +219,39 @@ def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
...
         try:
-            self._core_api.patch_node(
-                name=node_name,
-                body={
-                    "spec": {
-                        "unschedulable": True,
-                    },
-                },
+            ip = self._get_ip_from_node_name(node_name)
+            node = self._nodes_by_ip[ip] if ip in self._nodes_by_ip else None
```
piax93: This is the same as
```suggestion
            node = self._nodes_by_ip.get(ip)
```

(same below)

piax93 (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: get_node_by_hostname - done, I will keep taints part duplicate for now. I'll probably merge cordon/uncordon method.

PR #238, opened 2022-10-21
bumped sensu limit temporary.

ilkinmammadzada:
### Description

We are investigating slowness reason of metrics collector. 
We found some options to improve performance. 
But we should temporarily bump limit to make it less noisy.


BeryJu (approved):
I'm not 100% sure if this is enough to bump the alert timeout, I remember amjad was looking into this and had to increase a timeout in puppet (I think?) too

ilkinmammadzada:
@BeryJu Yeah, I searched key (check_clusterman_cluster_metrics_running) in sourcegraph and didn't find anything and thought that it should be enough (hopefully)

PR #239, opened 2022-10-24
optimized unschedulabe pod reason detection. 

ilkinmammadzada:
### Description

Currently we are checking nodes one by one if it has enough ClustermanResources (cpu, mem, disk, gpu) to decide unscheduling reason. We are stopping to iterate if we find any node which has more resource than needed for pod. But we are using lt (lower than) instead of le (lower than or equal) for comparison.
This means if you pod needs ClustermanResources(cpu=1, mem=1Gi, disk=10Gi, gpu=0) iteration won't be stopped if we don't have any node with GPU! 

I tested change performance with ~2500 pending pods. Running time was changed from 130 seconds to ~1 second. 

Additionally We probably should remove `_get_pod_unschedulable_reason` function because if pod has condition as below this means pod couldn't be scheduled due to "InsufficientResources". We don't need iterate all nodes for that, Kubernetes did it for us. 
Condition: type: PodScheduled, status: false, reason: Unschedulable

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -401,7 +401,7 @@ class KubernetesClusterConnector(ClusterConnector):
                 available_node_resources = total_node_resources(
                     node, self._excluded_pods_by_ip.get(node_ip, [])
                 ) - allocated_node_resources(pods_on_node)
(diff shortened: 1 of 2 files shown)


gmdfalk (approved):

gmdfalk:
Great find!

PR #249, opened 2022-11-03
handling multiple termination of same instance

ajayOO8:
### Description
Jira: [CLUSTERMAN-686](https://jira.yelpcorp.com/browse/CLUSTERMAN-686)
There are cases where clusterman could try to `terminate` an already `terminated` instance. AWS complains about these case if it has already removed those `terminated` instance from their console, which could lead cman to throw errors.

Now, before terminating Clusterman would check if the instance is not already terminated; then proceeds to `terminate` them.
### Testing Done
`make test`
Tested `terminate_instances_by_id()` .

### DIFF
--- a/clusterman/aws/aws_resource_group.py
+++ b/clusterman/aws/aws_resource_group.py
@@ -140,6 +140,13 @@ class AWSResourceGroup(ResourceGroup, metaclass=ABCMeta):
 
         instance_weights = {}
         for instance in ec2_describe_instances(instance_ids):
+            # skip if instance already terminated
+            if instance["State"].get("Name") in ["terminated", "shutting-down"]:
+                logger.warning(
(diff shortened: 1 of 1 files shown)


clusterman/aws/aws_resource_group.py
```diff
@@ -140,6 +140,13 @@ def terminate_instances_by_id(self, instance_ids: List[str], batch_size: int = 5
 
         instance_weights = {}
         for instance in ec2_describe_instances(instance_ids):
+            # skip if instance already terminated
+            if instance["State"].get("Name") == "terminated" or instance["State"].get("Name") == "shutting-down":
```
ilkinmammadzada: Nit: this can be changed with (if instance["State"].get("Name") in  {"terminated", "shutting-down"} to increase readability

ilkinmammadzada (approved):

PR #250, opened 2022-11-03
Optimized Kubernetes Metrics collection

ilkinmammadzada:
### Description

Currently we call `get_unschedulable_pods` 4 more times to calculate pending cpu, mem, disk, gpu.
We can keep total pending resources in memory and serve it with high performance.


mattmb (approved):
lgtm, I suspect you'll have some tests to clean up but the logic makes sense to me. The load_pods_info bit will definitely run each time that clusterman gathers the metrics right?

ilkinmammadzada:
> lgtm, I suspect you'll have some tests to clean up but the logic makes sense to me. The load_pods_info bit will definitely run each time that clusterman gathers the metrics right?


I fixed test. right! 
metric collector calls reload_state of pool_manager (and it call reload_state of kubernetes_cluster_connector)

https://github.com/Yelp/clusterman/blob/master/clusterman/batch/cluster_metrics_collector.py#L155

jfongatyelp (approved):

PR #253, opened 2022-11-07
ignored pending pod reason detection

ilkinmammadzada:
### Description

We noticed that autoscaling works weirdly (doesn’t react to pending pods) for some pools.
After investigation we saw that weirdness exists only in pools which use internal signals. (autoscale_signal.internal:true)
To explain briefly these pools are using PendingPodsSignal locally (in Autoscaler component) instead of getting information from DynamoDB (ExternalSignal)

- What is difference between them?

PendingPodsSignal: considers only unschedulable pods which has InsufficientResources reason.
ExternalSignal: considers all unschedulable pods.

- Why most times PendingPodsSignal doesn’t react to pending pods? / Why most time we have many pods with Unknown reason?

Because reason detection mechanism has bug: It checks all unschedulable to be sure that if it fits to any existing nodes in the pool.. It marks fitted unschedulable pod as Unknown.
This means if you have 500 pending pods (with 5 CPU request) and 100 nodes (99 with 0 empty CPU, 1 with 5 free CPU) then Clusterman thinks that all pending pods are pending for Unknown reason. And doesn’t scale up cluster…

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -197,11 +197,18 @@ class KubernetesClusterConnector(ClusterConnector):
         )
 
     def get_unschedulable_pods(
-        self,
+        self, detect_reason: Optional[bool] = True
     ) -> List[Tuple[KubernetesPod, PodUnschedulableReason]]:
         unschedulable_pods = []
         for pod in self._unschedulable_pods:
-            unschedulable_pods.append((pod, self._get_pod_unschedulable_reason(pod)))
+            unschedulable_pods.append(
+                (
+                    pod,
+                    self._get_pod_unschedulable_reason(pod)
+                    if detect_reason
+                    else PodUnschedulableReason.InsufficientResources,
+                )
+            )
         return unschedulable_pods
 
     def drain_node(self, node_name: str, disable_eviction: bool) -> bool:
--- a/clusterman/signals/pending_pods_signal.py
+++ b/clusterman/signals/pending_pods_signal.py
@@ -58,7 +58,9 @@ class PendingPodsSignal(Signal):
         retry_on_broken_pipe: bool = True,
     ) -> Union[SignalResourceRequest, List[KubernetesPod]]:
         allocated_resources = self.cluster_connector.get_cluster_allocated_resources()
-        pending_pods = self.cluster_connector.get_unschedulable_pods()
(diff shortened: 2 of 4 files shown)


clusterman/signals/pending_pods_signal.py
```diff
@@ -58,7 +58,9 @@ def evaluate(
         retry_on_broken_pipe: bool = True,
     ) -> Union[SignalResourceRequest, List[KubernetesPod]]:
         allocated_resources = self.cluster_connector.get_cluster_allocated_resources()
-        pending_pods = self.cluster_connector.get_unschedulable_pods()
+        pending_pods = self.cluster_connector.get_unschedulable_pods(
+            detect_reason=self.parameters.get("detect_reason", True)
```
ilkinmammadzada, line 62: I'm curious that if I add "detect_reason: false" to parameter in the config , this line will convert it to bool?!

ilkinmammadzada: Or should I convert it to bool?

piax93: YAML parsers do the boolean conversion for you

piax93 (approved):
I can say that the Python looks good to me, but honestly don't have enough insight into the signaling system to say whether this is a good or bad strategy.

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -197,11 +197,18 @@ def get_resource_excluded(self, resource_name: str) -> float:
         )
 
     def get_unschedulable_pods(
-        self,
+        self, detect_reason: Optional[bool] = True
```
nemacysts, line 200: suggestion: might be nice to be consistent with the function name and name this parameter `get_reason` 

that said, maybe we could even name this something like `get_detailed_reason`? or add a docstring to this function? as it is right now, this is a bit confusing to read since the function signature says that this will return `List[Tuple[KubernetesPod, PodUnschedulableReason]]`, but that's a bit at odds with the added parameter

nemacysts (approved):

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: Yeap, I agree with your point. I will delete PodUnschedulableReason and all related thing after migration (after enabling flag everywhere).

PR #255, opened 2022-11-08
Optimized Kubernetes Metrics collection #2

ilkinmammadzada:
### Description

2nd part of the optimisation https://github.com/Yelp/clusterman/pull/250
Currently we are checking same pods (running) 4 times to find allocated specific resource type (like: cpu, mem)

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -77,6 +77,7 @@ class KubernetesClusterConnector(ClusterConnector):
     _pods_by_ip: Mapping[str, List[KubernetesPod]]
     _label_selectors: List[str]
     _unschedulable_pods_resources: ClustermanResources
+    _allocated_pods_resources: ClustermanResources
 
     def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) -> None:
         super().__init__(cluster, pool)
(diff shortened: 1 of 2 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -472,12 +475,17 @@ def _get_nodes_by_ip(self) -> Mapping[str, KubernetesNode]:
     def _get_pods_info(
         self,
     ) -> Tuple[
-        Mapping[str, List[KubernetesPod]], List[KubernetesPod], Mapping[str, List[KubernetesPod]], ClustermanResources
+        Mapping[str, List[KubernetesPod]],
+        List[KubernetesPod],
+        Mapping[str, List[KubernetesPod]],
+        ClustermanResources,
+        ClustermanResources,
```
piax93, line 482: It might be the case to turn this into a namedtuple, as it's starting to get a bit difficult to read with all these returned things.

piax93 (approved):
Minor comment, looks alright to me otherwise.

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: I was thinking to replace `_get_pods_info` function with `_reload_workloads` (or something like this) and fill properties (like _unschedulable_pods, _pods_by_ip) inside it instead of returning. since it (`_get_pods_info`) is private method. What do you think?

PR #264, opened 2022-11-23
node-migration: add orphan_capacity_tollerance option

piax93:
There are some pools which are large, and change a lot pretty rapidly, which cause them to always have a good chunk of orphan nodes, making node-migration job think they are unhealthy, even though they are functioning just fine. This adds an option to loosen the capacity health-check as long as the raw number for node capacity is higher than the autoscaler target.

### DIFF
--- a/clusterman/autoscaler/pool_manager.py
+++ b/clusterman/autoscaler/pool_manager.py
@@ -598,9 +598,16 @@ class PoolManager:
             and node_metadata.instance.instance_id in self.resource_groups[node_metadata.instance.group_id].instance_ids
         )
 
-    def is_capacity_satisfied(self) -> bool:
-        """States whether current pool capacity is considered in line with expectation"""
-        return self.non_orphan_fulfilled_capacity >= self.target_capacity
+    def is_capacity_satisfied(self, orphan_capacity_tollerance: float = 0) -> bool:
+        """States whether current pool capacity is considered in line with expectation
+
+        :param float orphan_capacity_tollerance: acceptable ratio of orphan capacity to still consider check satisfied
+        """
+        target_capacity = self.target_capacity  # just saving a bit of CPU since it is a computed property
+        return self.non_orphan_fulfilled_capacity >= target_capacity or (
+            self.fulfilled_capacity >= target_capacity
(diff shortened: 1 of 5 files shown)


docs/source/node_migration.rst
```diff
@@ -50,6 +50,8 @@ The allowed values for the migration settings are as follows:
 
 * ``health_check_interval``: how much to wait between checks when monitoring pool health (2 minutes by default).
 
+* ``orphan_capacity_tollerance``: acceptable ratio of orphan capacity over target capacity to still consider the pool healthy (float, 0 by default).
```
ilkinmammadzada: What do you think about adding max value (0.2)?

clusterman/autoscaler/pool_manager.py
```diff
@@ -598,9 +598,16 @@ def is_node_still_in_pool(self, node_metadata: ClusterNodeMetadata) -> bool:
...
-        """States whether current pool capacity is considered in line with expectation"""
-        return self.non_orphan_fulfilled_capacity >= self.target_capacity
+    def is_capacity_satisfied(self, orphan_capacity_tollerance: float = 0) -> bool:
+        """States whether current pool capacity is considered in line with expectation
+
+        :param float orphan_capacity_tollerance: acceptable ratio of orphan capacity to still consider check satisfied
+        """
+        target_capacity = self.target_capacity  # just saving a bit of CPU since it is a computed property
+        return self.non_orphan_fulfilled_capacity >= target_capacity or (
+            self.fulfilled_capacity >= target_capacity
```
ilkinmammadzada, line 608: nit: (fulfilled_capacity - non_orphan_fulfilled_capacity)/fulfilled_capacity <= orphan_capacity_tollerance

`(fulfilled_capacity - non_orphan_fulfilled_capacity)/fulfilled_capacity` is ratio of the orphan instances. 
may it be easy to read?

ilkinmammadzada (approved):

clusterman/autoscaler/pool_manager.py
piax93: I did too much low level programming in uni not to favour multiplication over division when possible 😅

PR #270, opened 2022-11-30
removed use_labels_for_pods flag

ilkinmammadzada:
### Description

Cleanup after https://github.com/Yelp/clusterman/pull/177 
We rolled out pool label filtering for pods everywhere.
It's time to clean-up flag and unused methods.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -27,8 +27,6 @@ from kubernetes.client import V1beta1Eviction
 from kubernetes.client import V1DeleteOptions
 from kubernetes.client import V1ObjectMeta
 from kubernetes.client.models.v1_node import V1Node as KubernetesNode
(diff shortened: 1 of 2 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -93,10 +90,7 @@ def __init__(self, cluster: str, pool: Optional[str], init_crd: bool = False) ->
         self._label_selectors = []
         if self.pool:
             # TODO(CLUSTERMAN-659): Switch to using just pool_label_key once the new node labels are applied everywhere
```
piax93, line 92: The TODO comment should be removed

ilkinmammadzada: Nope actually, we are still using `node_label_key` for node filtering. Because we don't have new label in all of our nodes.

piax93 (approved):

PR #276, opened 2022-12-08
Remove exception when requesting a resource a pool doesn't have

nemacysts:
This is triggering when scaling a GPU-enabled pool from 0 since our bootstrap doesn't immediatly register these nodes as having any GPU capacity right now

### DIFF
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -30,7 +30,6 @@ from clusterman.autoscaler.config import get_autoscaling_config
 from clusterman.autoscaler.pool_manager import PoolManager
 from clusterman.config import POOL_NAMESPACE
 from clusterman.exceptions import NoSignalConfiguredException
-from clusterman.exceptions import ResourceRequestError
 from clusterman.interfaces.signal import Signal
 from clusterman.kubernetes.kubernetes_cluster_connector import KubernetesClusterConnector
 from clusterman.monitoring_lib import get_monitoring_client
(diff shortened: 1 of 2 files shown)


nemacysts:
we were seeing `FAILED: clusterman autoscaler failed (Signal requested 2 for gpus but the cluster doesn't have any of that resource)` internally when scaling gpu pools from 0 using the PendingPodsSignal

clusterman/autoscaler/autoscaler.py
```diff
@@ -399,9 +398,11 @@ def _get_most_constrained_resource_for_request(
 
             if resource_total == 0:
                 if resource_request_value > 0:
-                    raise ResourceRequestError(
+                    logger.warning(
                         f"Signal requested {resource_request_value} for {resource} "
-                        "but the cluster doesn't have any of that resource"
+                        "but the cluster doesn't have any of that resource - continuing "
+                        "under the assumption that the resource will be added soon (e.g., "
+                        "after a Puppet run"
```
jfongatyelp: typo: Missing closing `)` in this text

jfongatyelp (approved):
It seems that we only ever raised this as if someone began requesting GPUs on a non-gpu pool, this was the only way to let them know about it ([slack](https://yelp.slack.com/archives/CA6TN8J3G/p1562865244383400)).

Now since there's some delay in detecting gpus in these pools, this seems like the right change for now.

clusterman/autoscaler/autoscaler.py
nemacysts:
```suggestion
                        "after a Puppet run)"
```

ilkinmammadzada (approved):

PR #277, opened 2022-12-12
node-migration: add max_uptime_worker_skips setting

piax93:
Allow the uptime workers to eventually get the node draining going even when pool capacity is currently not fully satisfied.
It's just an in-memory value, so it'll get lost with restarts, but don't really see that as a bit issue.

### DIFF
--- a/clusterman/migration/worker.py
+++ b/clusterman/migration/worker.py
@@ -186,6 +186,7 @@ def uptime_migration_worker(
     :param int uptime_seconds: uptime threshold
     :param WorkerSetup worker_setup: migration setup
     """
+    skipped_executions = 0
     manager = PoolManager(cluster, pool, SUPPORTED_POOL_SCHEDULER, fetch_state=False)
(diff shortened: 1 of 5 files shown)


clusterman/migration/worker.py
```diff
@@ -194,10 +195,14 @@ def uptime_migration_worker(
     try:
         while True:
             manager.reload_state(load_pods_info=not worker_setup.ignore_pod_health)
-            if manager.is_capacity_satisfied(worker_setup.orphan_capacity_tollerance):
+            if (
+                worker_setup.max_uptime_worker_skips > 0 and skipped_executions > worker_setup.max_uptime_worker_skips
+            ) or manager.is_capacity_satisfied(worker_setup.orphan_capacity_tollerance):
```
piax93, line 200: `black` formatting horrifying as always ...

ilkinmammadzada (approved):

PR #281, opened 2022-12-15
added get_instance_type function

ajayOO8:
### Description
removed hardcoded variable clusterman.aws.markets.EC2_INSTANCE_TYPE with a more dynamic function `get_instance_type()`
added flag `clusters.{cluster_name}.enable_dynamic_instance_types` for controlling this change cluster wise.

### Testing done

Tested the individual function `get_instance_type()`  by hardcoding call to this function
`python3 -m clusterman.aws.markets`
executed `make test` and make itests`


.gitignore
```diff
@@ -16,6 +16,7 @@ precomputed
 *.sublime-*
 virtualenv_run
 .tox
+.toxb
```
ilkinmammadzada: Could you please explain this? :)

clusterman/aws/client.py
```diff
@@ -77,7 +77,7 @@ def _init_session():
         _session = boto3.session.Session(
             staticconf.read_string("accessKeyId", namespace=CREDENTIALS_NAMESPACE),
             staticconf.read_string("secretAccessKey", namespace=CREDENTIALS_NAMESPACE),
-            region_name=staticconf.read_string("aws.region"),
+            region_name="us-east-1",
```
ilkinmammadzada: this look dangerous. We use this module for other purpose as well.

clusterman/util.py
```diff
@@ -74,13 +74,16 @@ class ClustermanResources(NamedTuple):
...
+        resources = get_instance_type(instance_type)
+        if resources is not None:
+            return ClustermanResources(
+                cpus=resources.cpus,
+                mem=resources.mem * 1024,  # AWS metadata for RAM is in GB
+                disk=(resources.disk or DEFAULT_VOLUME_SIZE_GB) * 1024,  # AWS metadata for disk is in GB
+                gpus=resources.gpus,
+            )
+        else:
+            return ClustermanResources()
```
ilkinmammadzada: What is the current behaviour in case? I mean what happens if receive non-exist `instance_type` ?

tests/simulator/simulated_cluster_connector_test.py
```diff
@@ -26,7 +27,18 @@
 from clusterman.simulator.simulated_spot_fleet_resource_group import SimulatedSpotFleetResourceGroup
 
 
-TEST_MARKET = InstanceMarket("c3.4xlarge", "us-west-2a")
+
```
ilkinmammadzada: should we mock `get_instance_type` ?

ilkinmammadzada (commented):
some comments

clusterman/util.py
ajayOO8: it probably would raise KeyERROR exception.

.gitignore
ajayOO8: .toxb is the workspace for tox python

clusterman/aws/client.py
ajayOO8: it was for local testing, will remove this in Final PR

clusterman/util.py
ilkinmammadzada: Yeah, I mean we should try to keep same behaviour in our change.

ilkinmammadzada:
How do you plan to rollout change gradually? @ajayOO8

clusterman/util.py
ajayOO8: yeah sure.

ajayOO8:
> How do you plan to rollout change gradually? @ajayOO8

I was thinking to roll it out in lower environment only and observe for few days, but since all of our environment uses common code base,, it would be really tricky to rollout to only one env. 
May be if I can pin a specific version of Clusterman to lower env then i can test out this feature.

ilkinmammadzada:
Unfortunately, it's hard to pin specific version. So it would be better to have feature flag to rollout easily.

clusterman/util.py
```diff
@@ -74,13 +74,16 @@ class ClustermanResources(NamedTuple):
...
+        resources = get_instance_type(instance_type)
+        if resources is not None:
+            return ClustermanResources(
+                cpus=resources.cpus,
+                mem=resources.mem * 1024,  # AWS metadata for RAM is in GB
+                disk=(resources.disk or DEFAULT_VOLUME_SIZE_GB) * 1024,  # AWS metadata for disk is in GB
+                gpus=resources.gpus,
+            )
+        else:
+            raise Exception(f"instance type {instance_type} is invalid.")
```
ilkinmammadzada: maybe we should move non-exis/invalid exception to get_instance_type function. And it would help us to keep other behaviours same as before. 

- get_instance_type replaced EC2_INSTANCE_TYPES map. we got exception if we request non-exist type. 
- get_market_resources returned InstanceResources, but now it's optional.

tests/simulator/simulated_cluster_connector_test.py
```diff
@@ -16,17 +16,20 @@
...
 from clusterman.interfaces.types import AgentState
 from clusterman.interfaces.types import ClustermanResources
+from clusterman.simulator.simulate_aws_market import mock_get_market_resources
+from clusterman.simulator.simulate_aws_market import simulate_InstanceMarket
 from clusterman.simulator.simulated_aws_cluster import Instance
 from clusterman.simulator.simulated_cluster_connector import SimulatedClusterConnector
 from clusterman.simulator.simulated_spot_fleet_resource_group import SimulatedSpotFleetResourceGroup
 
+# from clusterman.aws.markets import get_market_resources
+# from clusterman.aws.markets import InstanceMarket
```
ilkinmammadzada: we should remove unnecessary imports

clusterman/simulator/simulate_aws_market.py
```diff
@@ -0,0 +1,62 @@
...
+    "us-west-1a",
+    "us-west-1b",
+    "us-west-1c",
+    "us-west-2a",
+    "us-west-2b",
+    "us-west-2c",
+]
+
+
+class simulate_InstanceMarket(_InstanceMarket):
```
ilkinmammadzada: Could you please describe briefly reason of creating this class?

clusterman/aws/markets.py
```diff
@@ -379,8 +381,38 @@ def parse(cls, string: str):
 ]
 
 
+@lru_cache(maxsize=128)
+def fetch_instance_type_from_aws(instance_type: str) -> InstanceResources:
+    res = {}
+    try:
+        res = ec2.describe_instance_types(InstanceTypes=[instance_type]).get("InstanceTypes")[0]
+    except Exception as e:
+        raise ValueError(f"Error occoured while describing instance type {instance_type} : {e}")
```
ilkinmammadzada, line 390: will this run if we send request for non-exist type on line 388?

ilkinmammadzada (approved):
LGTM! only one question

PR #285, opened 2023-01-04
[POC] node-migration: apply prescaling as temporary offset

piax93:
This turns the pre-scaling setting into a temporary offset which is always applied on top of the calculated pool capacity


clusterman/migration/worker.py
```diff
@@ -259,9 +261,18 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
             nodes = manager.get_node_metadatas(AWS_RUNNING_STATES)
             offset = worker_setup.prescaling.of(len(nodes))
             logger.info(f"Applying pre-scaling of {offset} node to {migration_event.cluster}:{migration_event.pool}")
-            avg_weight = mean(node.instance.weight for node in nodes)
-            prescaled_capacity = round(manager.target_capacity + (offset * avg_weight))
-            manager.modify_target_capacity(prescaled_capacity)
+            capacity_offset = offset * mean(node.instance.weight for node in nodes)
+            if worker_setup.disable_autoscaling:
```
ilkinmammadzada, line 265: Do we still need `disable_autoscaling` setting? Because it actually doesn't disable autoscaling.

clusterman/migration/worker.py
```diff
@@ -283,8 +294,14 @@ def event_migration_worker(migration_event: MigrationEvent, worker_setup: Worker
...
-        # trust the autoscaler to readjust that in a short time eventually
+        # we do not reset the pool target capacity in case of direct capacity changes
+        # as we trust the autoscaler to readjust that in a short time eventually
         if worker_setup.disable_autoscaling:
             logger.info(f"Re-enabling autoscaling for {migration_event.cluster}:{migration_event.pool}")
             enable_autoscaling(migration_event.cluster, migration_event.pool, SUPPORTED_POOL_SCHEDULER)
+        elif worker_setup.prescaling:
+            remove_capacity_offset(
+                migration_event.cluster,
+                migration_event.pool,
```
ilkinmammadzada, line 305: We are using cluster, pool and scheduler as a key in some places. Should we check (and reject) if requested pool already has on-going migration?

ilkinmammadzada (commented):
Looks very good generally! Some small questions. 

and one unrelated comment: Can we add output (something like "request received") for `clusterman migrate` command?

clusterman/migration/worker.py
piax93: It does on line 252. Mostly keeping it for smoother transition since we have a few thing configured that way, so
* disable autoscaling + prescaling does the old thing
* just prescaling uses the new logic

Then I can work on removing the settings gradually.

clusterman/migration/worker.py
piax93: It's impossible for a pool to have concurrent migrations running, there's a lock preventing that.

piax93: Rather, "impossible as long as you don't run more than one instance of the batch per cluster", which must not be done as many more things would break.

clusterman/migration/worker.py
ilkinmammadzada: oops, Actually I asked question on line 252 :)
Cool, I loved idea.

clusterman/migration/worker.py
ilkinmammadzada: As I understood we are checking it while worker running. I meant directly decline 2nd migration creation and respond to client via CLI. 
We are deleted (actually changing label to `STOP`) pending and in-progress migration events when we run `migrate-stop` command. This is a bit implicit.
But having multiple migration event and processing them one by one is cool.

clusterman/autoscaler/offset.py
```diff
@@ -0,0 +1,94 @@
...
+    dynamodb.delete_item(
+        TableName=staticconf.read("aws.state_table", default=CLUSTERMAN_STATE_TABLE),
+        Key={
+            "state": {"S": AUTOSCALER_CAPACITY_OFFSET_KEY},
+            "entity": {"S": f"{cluster}.{pool}.{scheduler}"},
+        },
+    )
+
+
+def get_capacity_offset(cluster: str, pool: str, scheduler: str, timestamp: arrow.Arrow) -> float:
```
ilkinmammadzada, line 69: Should we handle exception and return 0 in case of it occurs? Because it's not really necessary for autoscaling.

ilkinmammadzada (approved):
lgtm

clusterman/autoscaler/offset.py
piax93: Good call. I kind of assumed that the dynamodb wrapper does this already, since this method is basically a copy-pasta of what's used for toggling the autoscaler, but let me double check.

piax93: Looks like it does not. Likely dynamodb is so reliable that we never encountered a read error, but in this case it definitely makes sense to handle it.

PR #290, opened 2023-01-12
U/jfong/emit more autoscaler metrics

jfongatyelp:
### Description

This adds both max_capacity and setpoint to the `clusterman.autoscaler` metrics emitted by the autoscaler, for use in alerting/graphs e.g. when a pool has been at max_capacity or over setpoint for some period of time.

Note that I opted to emit these alongside the other metrics we emit from the autoscaler like target_capacity since we don't have access to the autoscalerconfig from the poolmanager, which would be the only thing we have access to from the cluster_metrics branch when emitting metadata metrics. 

Unfortunately it looks like the autoscaler-emitted metrics never actually got updated to include scheduler as part of the dimensions (so a mesos 'default' pool would emit the exact same metric name+dimensions as a kubernetes 'default' pool), so this PR corrects that problem as well. 
### Testing Done

Added one unit test to verify we're setting these values as expected, and manually applied to our kubestage autoscaler and verified the metrics were being emitted as expected: 
Screen Shot 2023-01-12 at 12 49 48 PM

### DIFF
--- a/clusterman/autoscaler/autoscaler.py
+++ b/clusterman/autoscaler/autoscaler.py
@@ -45,6 +45,8 @@ from clusterman.util import Status
 
 SIGNAL_LOAD_CHECK_NAME = "signal_configuration_failed"
 TARGET_CAPACITY_GAUGE_NAME = "clusterman.autoscaler.target_capacity"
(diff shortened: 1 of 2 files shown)


nemacysts (approved):

ilkinmammadzada (approved):
LGTM! Thanks for switching to this option :)

PR #292, opened 2023-02-02
Add r6idn instances

sidsatyelp:
Based on the issue that flink-spot pool is seeing pending pods

https://yelp.slack.com/archives/C01S98B4946/p1675363606557849


nemacysts (approved):
looking forward to the day we no longer need to do this :p

sofyat (approved):

PR #295, opened 2023-02-09
Making spot instance warning processing configurable

ilkinmammadzada:
### Description

Some teams don't need cleaning up 2 minutes before real termination. Therefore we made it configurable.


nemacysts (approved):
seems fine to me :)

88manpreet (approved):
Thanks for making this change.

ankit864 (approved):

PR #297, opened 2023-02-10
node-migration: offset uptime checks by event creation time

piax93:
Sibling of https://github.com/Yelp/clusterman/pull/296.

This adds an offset based on the creation time of the migration event when doing checks on host uptime, so that if a job were to be restarted let's say 2 days after its initial definition, with a condition like `uptime lt 1d`, instances which were started during that timespan would not be matched again, as the condition would have implicitly become `uptime lt (1+2)d`.


ilkinmammadzada (approved):
great idea!

PR #299, opened 2023-02-13
refactoring CMAN_CLUSTER var

ajayOO8:
[DRAFT]
refactoring CMAN_CLUSTER variable

### DIFF
--- a/clusterman/aws/auto_scaling_resource_group.py
+++ b/clusterman/aws/auto_scaling_resource_group.py
@@ -153,6 +153,7 @@ class AutoScalingResourceGroup(AWSResourceGroup):
             options.extend(
                 self._get_options_for_instance_type(
                     override["InstanceType"],
+                    self.cluster,
                     float(override["WeightedCapacity"]),
                 )
             )
@@ -161,7 +162,7 @@ class AutoScalingResourceGroup(AWSResourceGroup):
         if not options:
             options.extend(
                 self._get_options_for_instance_type(
-                    self._launch_template_config["LaunchTemplateData"]["InstanceType"],
+                    self._launch_template_config["LaunchTemplateData"]["InstanceType"], self.cluster
                 )
             )
 
@@ -234,6 +235,7 @@ class AutoScalingResourceGroup(AWSResourceGroup):
     def _get_options_for_instance_type(
         self,
         instance_type: str,
+        cluster: str,
         weight: Optional[float] = None,
     ) -> List[ClusterNodeMetadata]:
         """Generate a list of possible ClusterNode types that could be added to this ASG,
@@ -246,7 +248,7 @@ class AutoScalingResourceGroup(AWSResourceGroup):
             weight = weight or self.market_weight(instance_market)
             options.append(
                 ClusterNodeMetadata(
-                    agent=AgentMetadata(total_resources=ClustermanResources.from_instance_type(instance_type)),
+                    agent=AgentMetadata(total_resources=ClustermanResources.from_instance_type(instance_type, cluster)),
                     instance=InstanceMetadata(market=instance_market, weight=weight),
                 )
             )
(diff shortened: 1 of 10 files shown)


clusterman/aws/auto_scaling_resource_group.py
```diff
@@ -246,7 +248,7 @@ def _get_options_for_instance_type(
             weight = weight or self.market_weight(instance_market)
             options.append(
                 ClusterNodeMetadata(
-                    agent=AgentMetadata(total_resources=ClustermanResources.from_instance_type(instance_type)),
+                    agent=AgentMetadata(total_resources=ClustermanResources.from_instance_type(instance_type, cluster)),
```
piax93: No need to have cluster passed as a function argument, you can read it as a property from `self` directly.

clusterman/aws/aws_resource_group.py
```diff
@@ -284,7 +284,7 @@ def load(  # type: ignore # (mypy errors with "incompatible signature with super
                 if tag_json:
                     identifier_tags = json.loads(tag_json)
                     if identifier_tags["pool"] == pool and identifier_tags["paasta_cluster"] == cluster:
-                        rg = cls(rg_id, **kwargs)
+                        rg = cls(cluster, rg_id, **kwargs)
```
piax93: `cluster` would be the second positional argument, no the first

clusterman/aws/markets.py
```diff
@@ -417,8 +417,8 @@ def fetch_instance_type_from_aws(instance_type: str) -> InstanceResources:
     return InstanceResources(vcpu_count, mem_size, disk_size, gpu_size)
 
 
-def get_instance_type(instance_type: str) -> InstanceResources:
-    cluster_name = os.environ.get("CMAN_CLUSTER", None)
+def get_instance_type(instance_type: str, cluster: str = None) -> InstanceResources:
+    cluster_name = os.environ.get("PAASTA_CLUSTER", cluster)
```
piax93: I'd suggest to just make the function argument compulsory and authoritative to avoid surprises

```suggestion
def get_instance_type(instance_type: str, cluster_name: str) -> InstanceResources:
```

clusterman/aws/spot_fleet_resource_group.py
```diff
@@ -226,6 +229,6 @@ def load_spot_fleets_from_s3(bucket: str, prefix: str, pool: str = None) -> Mapp
             if pool and resource["pool"] != pool:
                 continue
 
-            spot_fleets[resource["id"]] = SpotFleetResourceGroup(resource["id"])
+            spot_fleets[resource["id"]] = SpotFleetResourceGroup(cluster, resource["id"])
```
piax93: Same argument order issue as above

clusterman/draining/queue.py
```diff
@@ -542,7 +542,7 @@ def terminate_host(self, host: Host) -> None:
             resource_group._reload_resource_group()
         else:
             resource_group_class = RESOURCE_GROUPS[host.sender]
-            resource_group = resource_group_class(host.group_id)
+            resource_group = resource_group_class(self.cluster, host.group_id)
```
piax93: Here as well

piax93 (changes requested):
There's also this bit which uses the function
https://github.com/Yelp/clusterman/blob/master/clusterman/migration/event.py#L64-L67 , but considering it's just some "excess of validation" I'd be happy to give it up, getting the `cluster` variable in there wouldn't be the easiest.
Or we can move the validation downstream, in the event worker implementation.

clusterman/aws/markets.py
ajayOO8: making `cluster` var non-default is tricky, a lot for mock test depends on this and InstanceMarket.
@piax93

ajayOO8:
opened #300
closing this.

PR #301, opened 2023-02-14
Make pending pods signal much more aggressive.

bobtfish:
This should force a scale up if there are pending pods, which will hopefully help get us unstuck with binpacking and Flink workloads.


piax93 (approved):
"2" looked already quite arbitrary here, so using a different arbitrary value should be fine

marcos-sb (approved):

mattmb:
I'd prefer if we can make it a config value (srvconfigs is easier to roll forward back too).

EvanKrall:
Can you provide some context (a ticket number, slack link, or something) as to why you made this change?

bobtfish:
Yep, my bad. This was post discussion here: https://yelp.slack.com/archives/CA6TN8J3G/p1676386279337199 and ticket is [CLUSTERMAN-779](https://jira.yelpcorp.com/browse/CLUSTERMAN-779)

PR #302, opened 2023-02-15
Move get resource request

bobtfish:
### Description

Revert the change to the multiplier for now, and make it configurable per pool as per @mattmb s request.

This will allow us to turn it up for Flink pools independently of anything else for testing.

Also add log line which will tell us how many resources the `PendingPodsSignal` is adding, to allow us to work out what a correct multiplier should actually look like, and generally better work out what is going on.

### Testing Done

Added unit tests for the new config value.

### Related context
https://yelp.slack.com/archives/CA6TN8J3G/p1676386279337199
[CLUSTERMAN-779](https://jira.yelpcorp.com/browse/CLUSTERMAN-779)


piax93 (approved):
🚢

PR #305, opened 2023-02-23
Created version 2 for pending pods signal

ilkinmammadzada:
### Description

Clusterman doesn't bump capacity always when we have pending pods. It may blocked by target_capacity_margin.
It's also possible that pending pods signal requests resource less then current resource even if we have pending pods in the pool.

### Solution

Creating a new logic that always bump capacity if we have single pending pods in the pool


bobtfish (commented):
this lgtm, but being clear here - this will bump the capacity every autoscaler run whilst we have pending pods, right?

Which means that if we break things so that new capacity can't come online, we'll scale up to the max allowed capacity.

Do I have that right?

clusterman/signals/pending_pods_signal.py
```diff
@@ -65,3 +76,20 @@ def _get_resource_request(
...
+        target_capacity_margin: float,
+        pending_pods: Optional[List[KubernetesPod]] = None,
+    ) -> SignalResourceRequest:
+        """Given a list of metrics, construct a resource request based on the most recent
+        data for allocated and pending pods"""
+
+        pending_pods = pending_pods or []
+
+        if len(pending_pods) > 0:
+            return total_resources * (1 + target_capacity_margin)
```
ilkinmammadzada: we need better calculation here. pending pods may request more than margin

tub: What are we trying to achieve here? We want clusterman to add new hosts whenever there's pending pods right?

How about if we added a fixed amount of resource request, enough to scale up by at least N nodes in the pool. This would prevent overscaling like we had yesterday but also guarantee that we scale up _by something_ whenever there's pending pods?

tub: (Bear in mind this is quite possibly a terrible idea, I've not had coffee yet)

clusterman/signals/pending_pods_signal.py
```diff
@@ -41,6 +43,15 @@ def evaluate(
         # Get the most recent metrics _now_ and when the boost was set (if any) and merge them
         if self.parameters.get("per_pod_resource_requests"):
             return pending_pods
+        elif self.parameters.get("v2", True):
+            total_resources = self.cluster_connector.get_cluster_total_resources()
```
ilkinmammadzada, line 47: I need to replace it with currenct_target_capacity

bobtfish (changes requested):
We presumably want to add some tests for this? Looks reasonable though

bobtfish (approved):

PR #307, opened 2023-02-24
Changed excluded resource (daemonset) logic

ilkinmammadzada:
### Description

After moving pool label based filtering for pods (CLUSTERMAN-582) we lost excluding daemonset resource feature.
Because daemonsets don't have pool label, which is expected.
We are going to move single excluded resource logic for the pool, because we have same daemonset pods on all node in the same pool.

### DIFF
--- a/tests/kubernetes/kubernetes_cluster_connector_test.py
+++ b/tests/kubernetes/kubernetes_cluster_connector_test.py
@@ -384,9 +384,8 @@ def mock_cluster_connector_crd(mock_cluster_connector):
 )
 def test_get_agent_metadata(mock_cluster_connector, ip_address, expected_state):
     with PatchConfiguration(
-            {"exclude_daemonset_pods": True},
-            namespace=POOL_NAMESPACE.format(pool=mock_cluster_connector.pool,
-                                            scheduler=mock_cluster_connector.SCHEDULER),
+        {"exclude_daemonset_pods": True},
+        namespace=POOL_NAMESPACE.format(pool=mock_cluster_connector.pool, scheduler=mock_cluster_connector.SCHEDULER),
     ):
         mock_cluster_connector.reload_state()
         agent_metadata = mock_cluster_connector.get_agent_metadata(ip_address)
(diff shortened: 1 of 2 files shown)


piax93:
Implementation seems sensible to me, but tests need fixing.

ilkinmammadzada:
@piax93 it should be fine now.

tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -156,7 +156,30 @@ def pending_pod():
 
 
 @pytest.fixture
-def daemonset_pod():
+def daemonset_pod_1():
```
ilkinmammadzada, line 159: added daemonset pod for each node

tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -178,6 +201,29 @@ def daemonset_pod():
     )
 
 
+@pytest.fixture
+def daemonset_pod_3():
```
ilkinmammadzada, line 205: added daemonset pod for each node

tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -333,9 +383,14 @@ def mock_cluster_connector_crd(mock_cluster_connector):
...
 )
 def test_get_agent_metadata(mock_cluster_connector, ip_address, expected_state):
-    agent_metadata = mock_cluster_connector.get_agent_metadata(ip_address)
-    assert agent_metadata.is_safe_to_kill == (ip_address != "10.10.10.2")
-    assert agent_metadata.state == expected_state
+    with PatchConfiguration(
+        {"exclude_daemonset_pods": True},
+        namespace=POOL_NAMESPACE.format(pool=mock_cluster_connector.pool, scheduler=mock_cluster_connector.SCHEDULER),
+    ):
+        mock_cluster_connector.reload_state()
```
ilkinmammadzada, line 390: enabled exclude_daemonset_pods to keep one node `idle`

tests/kubernetes/kubernetes_cluster_connector_test.py
```diff
@@ -353,17 +408,23 @@ def test_reload_state_no_pods(mock_cluster_connector):
 
 
 def test_allocation(mock_cluster_connector):
-    assert mock_cluster_connector.get_resource_allocation("cpus") == 7.5
+    assert mock_cluster_connector.get_resource_allocation("cpus") == 10.5
```
ilkinmammadzada, line 411: Added two ds pod, this means +3 vCPUs

piax93 (approved):

PR #308, opened 2023-03-03
Better prioritization for choosing nodes as victim for scaling down

ilkinmammadzada:
### Description

Default pool case:
They have nrt search pods (which request bigger resources, and mostly we have only 1 or 2 nrt pods in our nodes)
Clusterman chooses nodes which has less pods for termination. This caused many scale up/down in mentioned pools.

- Clusterman sees pending pods, bumps capacity
- Nrt pods are scheduled
- Clusterman wants to scale down, because pending pods signal v2 bumps a bit more as expected.
- Clusterman delete nodes which have nrt (because they have small amount pods)

And again and again. 
I’m expecting this will support for better bin-packing as well.

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -501,7 +501,7 @@ class KubernetesClusterConnector(ClusterConnector):
 
     def get_node_priority(self, node_id: str) -> float:
         pods = self._pods_by_ip[node_id]
-        if not pods or len(pods) == 0:
+        if not pods:
             return 0.0
         max_requested_resource = max([self.get_pod_resources_score(pod) for pod in pods])
         # nodes have bigger pods = higher priority = lower possibility for choosing termination
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -383,6 +383,7 @@ class KubernetesClusterConnector(ClusterConnector):
             is_safe_to_kill=self._is_node_safe_to_kill(node_ip),
             is_draining=self._is_node_draining(node),
             batch_task_count=self._count_batch_tasks(node_ip),
+            priority=self.get_node_priority(node_ip),
             state=(AgentState.RUNNING if self._pods_by_ip[node_ip] else AgentState.IDLE),
             task_count=len(self._pods_by_ip[node_ip]),
             total_resources=total_node_resources(node, self._excluded_pods),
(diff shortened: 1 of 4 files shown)


clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -498,6 +499,20 @@ def _is_unschedulable(self, pod: KubernetesPod) -> bool:
                 return True
         return False
 
+    def get_node_priority(self, node_id: str) -> float:
+        pods = self._pods_by_ip[node_id]
+        if not pods or len(pods) == 0:
```
gmdfalk: i'm pretty sure `not pods` covers both nil and empty so you could omit `or len(pods)`

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -498,6 +499,20 @@ def _is_unschedulable(self, pod: KubernetesPod) -> bool:
...
+            return 0.0
+        max_requested_resource = max([self.get_pod_resources_score(pod) for pod in pods])
+        # nodes have bigger pods = higher priority = lower possibility for choosing termination
+        return max_requested_resource
+
+    def get_pod_resources_score(self, pod: KubernetesPod) -> float:
+        resource = total_pod_resources(pod)
+        cpus = getattr(resource, "cpus") * 2
+        mem = getattr(resource, "mem") / 1000
```
gmdfalk, line 513: is there any possibility of unit confusion here with human readable unit input like 1 MiB etc?

clusterman/simulator/simulated_pool_manager.py
```diff
@@ -58,6 +58,9 @@ def __init__(
             self.pool_config.read_int("scaling_limits.min_node_scalein_uptime_seconds", default=-1),
             MAX_MIN_NODE_SCALEIN_UPTIME_SECONDS,
         )
+        self.killable_nodes_prioritizing_v2 = self.pool_config.read_bool(
```
gmdfalk, line 61: how is this simulated? can we have a test?

clusterman/kubernetes/kubernetes_cluster_connector.py
```diff
@@ -498,6 +499,20 @@ def _is_unschedulable(self, pod: KubernetesPod) -> bool:
                 return True
         return False
 
+    def get_node_priority(self, node_id: str) -> float:
+        pods = self._pods_by_ip[node_id]
+        if not pods or len(pods) == 0:
+            return 0.0
+        max_requested_resource = max([self.get_pod_resources_score(pod) for pod in pods])
```
gmdfalk, line 506: why use only the single biggest pod to gauge priority? why not e.g. calculate the priority as a product of how many pods the node has (more disruption) and their average size (more difficult to reschedule)?

ilkinmammadzada: I was thinking about average/max/Percentile XX and etc. Actually none of them can solve problem totally. I just wanted to do an experiment with it. 
I'm currently doing some science: 
* Marking node killable only if its existing pods can be scheduled to other nodes.
Which is really not easy with our current architecture

clusterman/kubernetes/kubernetes_cluster_connector.py
ilkinmammadzada: We are keeping mem with MB only.

gmdfalk (approved):

PR #310, opened 2023-03-08
Added details to the log for marked nodes.

ilkinmammadzada:
### Description

Added more details to the log about nodes which marked for draining.


nemacysts (approved):
more info++

PR #312, opened 2023-03-10
Ignoring expired pending pods

ilkinmammadzada:
### Description

We want to have limit for pending pods. If pending pods exist more than configured time, we will ignore them to calculate needed resources. Because probably they misconfigured.

### DIFF
--- a/clusterman/signals/pending_pods_signal.py
+++ b/clusterman/signals/pending_pods_signal.py
@@ -94,6 +94,10 @@ class PendingPodsSignal(Signal):
         pending_pods = pending_pods or []
         filtered_pending_pods = [pod for pod in pending_pods if not self._ignore_peding_pod(pod)]
 
+        ignored_pending_pods_count = len(pending_pods) - len(filtered_pending_pods)
+        if ignored_pending_pods_count > 0:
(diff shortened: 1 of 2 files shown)


ajayOO8 (approved):

clusterman/signals/pending_pods_signal.py
```diff
@@ -91,9 +92,14 @@ def _get_resource_request_v2(
...
         resource_request = SignalResourceRequest()
         pending_pods = pending_pods or []
+        filtered_pending_pods = [pod for pod in pending_pods if not self._ignore_peding_pod(pod)]
 
-        if len(pending_pods) > 0:
-            for pod in pending_pods:
+        ignored_pending_pods_count = len(pending_pods) - len(filtered_pending_pods)
+        if ignored_pending_pods_count > 0:
+            logger.info(f"Ignoring {ignored_pending_pods_count} Pending pods due to max_scheduling_seconds")
+
```
ankit864, line 100: do filtered pending pod have any action after they are filtered as in race condition where we are not increasing capacity and eventually the reason is "not enough resources for the pods"?

ilkinmammadzada: As we discussed we will set bigger limit like more than 5 hours.

PR #315, opened 2023-04-10
Allow overriding srv-configs-provided kubeconfig

nemacysts:
This serves two purposes:
* we can get rid of SSH access to our kube masters (and the sudoers
  rules for running clusterman) in favor of using the local
  PaaSTA-provided kubeconfig and aws-okta auth
* allows us to run the clusterman CLI for EKS pools (where there are no
  masters to SSH into)


nemacysts:
(this needs to be shipped before https://github.yelpcorp.com/sysgit/puppet/pull/9189)

jfongatyelp (approved):

piax93 (approved):
++

PR #316, opened 2023-04-12
Allow setting an AWS profile override (+ support pod identity)

nemacysts:
Right now, Clusterman will always use /etc/boto_cfg creds - which means that we will not be able to access prod resources from devboxen normally.

We can do one of two things:
* Use our personal permissions
* Route folks to boxes like adhoc-prod for running clusterman against prod clusters

We'd like to minimize our usage of adhoc-prod in general, so I've gone for the first option and will be setting the relevant env vars in our wrappers.

Additionally, this PR allows us to use Pod Identity if the required environment variables are present

### DIFF
--- a/clusterman/aws/client.py
+++ b/clusterman/aws/client.py
@@ -29,6 +29,7 @@ from mypy_extensions import TypedDict
 from retry import retry
 
 from clusterman.config import CREDENTIALS_NAMESPACE
+from clusterman.exceptions import InvalidConfigurationError
 
 logger = colorlog.getLogger(__name__)
 _session = None
@@ -74,11 +75,21 @@ def _init_session():
     global _session
 
     if not _session:
-        _session = boto3.session.Session(
-            staticconf.read_string("accessKeyId", namespace=CREDENTIALS_NAMESPACE),
-            staticconf.read_string("secretAccessKey", namespace=CREDENTIALS_NAMESPACE),
-            region_name=staticconf.read_string("aws.region"),
-        )
+        # when running clusterman's CLI, we want to use folks' personal AWS creds
+        # so that we don't need to distribute service user creds (and since we can't
+        # use Pod Identity outside of k8s :p)
+        if os.getenv("AWS_PROFILE"):
+            if os.getenv("AWS_REGION") is None:
+                raise InvalidConfigurationError("Setting AWS_REGION is required if setting a specific AWS profile")
+            _session = boto3.session.Session(
+                region_name=os.getenv("AWS_REGION"),
+            )
+        else:
+            _session = boto3.session.Session(
+                staticconf.read_string("accessKeyId", namespace=CREDENTIALS_NAMESPACE),
+                staticconf.read_string("secretAccessKey", namespace=CREDENTIALS_NAMESPACE),
+                region_name=staticconf.read_string("aws.region"),
+            )
 
 
 def get_main_module_name() -> str:
--- a/clusterman/aws/client.py
+++ b/clusterman/aws/client.py
@@ -86,14 +86,6 @@ def _init_session():
             _session = boto3.session.Session(
                 region_name=staticconf.read_string("aws.region"),
             )
-        # this should only be used when running clusterman in paasta using pod identity
-        elif os.getenv("AWS_WEB_IDENTITY_TOKEN_FILE"):
-            _session = boto3.session.Session(
-                # unfortunately, it doesn't seem like the pod identity webhook lets us magically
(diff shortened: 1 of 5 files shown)


clusterman/run.py
```diff
@@ -27,16 +27,7 @@ def main(argv=None):
...
-        args.entrypoint(args)
-    except Exception as e:
-        print(f"Exception of type {e.__class__.__name__} occured")
-
-        if e.args:
-            for arg in e.args:
-                print(arg)
-
-        exit(1)
+    args.entrypoint(args)
```
nemacysts, line 30: i can revert this if we want - but hiding the traceback makes debugging issues somewhat annoying imo

tox.ini
```diff
@@ -7,11 +7,16 @@ tox_pip_extensions_ext_venv_update = true
 passenv = HOME SSH_AUTH_SOCK USER LANG PIP_INDEX_URL
 basepython = python3.7
 envdir = virtualenv_run
+usedevelop = true
```
nemacysts, line 10: this allows us to use the clusterman code in our venvs when running `clusterman` in our shells

clusterman/cli/util.py
```diff
@@ -21,16 +21,15 @@
...
 
 def timeout_wrapper(main):
     def wrapper(args: argparse.Namespace):
-        # After 10s, prints a warning message if the command is running from the wrong place
         def timeout_handler():
-            warning_string = "This command is taking a long time to run; are you running from the right account?"
+            warning_string = "This command is taking a long time to run; you're likely targetting a large pool/cluster."
             if "yelpcorp" in socket.getfqdn():
-                warning_string += "\nHINT: try ssh'ing to adhoc-prod or another box in the right account."
+                warning_string += "\nIf this command hasn't returned in several minutes, reach out to #clusterman"
```
nemacysts, line 32: this was my original change before i realized what was causing me to see weird behavior locally :p

clusterman/aws/client.py
```diff
@@ -74,11 +75,21 @@ def _init_session():
...
-            region_name=staticconf.read_string("aws.region"),
-        )
+        # when running clusterman's CLI, we want to use folks' personal AWS creds
+        # so that we don't need to distribute service user creds (and since we can't
+        # use Pod Identity outside of k8s :p)
+        if os.getenv("AWS_PROFILE"):
+            if os.getenv("AWS_REGION") is None:
+                raise InvalidConfigurationError("Setting AWS_REGION is required if setting a specific AWS profile")
+            _session = boto3.session.Session(
+                region_name=os.getenv("AWS_REGION"),
```
jfongatyelp: Hmm, interesting that we set this at the session level...I thought technically we can set this later when we generate the individual clients, and at that point we _should_ have access to the cluster config's `aws_region` setting.
But not a blocker at all, I know we can set this in the wrappers (or yelsoa-configs for instances) to sidestep!

clusterman/aws/client.py
```diff
@@ -74,11 +75,21 @@ def _init_session():
...
     if not _session:
-        _session = boto3.session.Session(
-            staticconf.read_string("accessKeyId", namespace=CREDENTIALS_NAMESPACE),
-            staticconf.read_string("secretAccessKey", namespace=CREDENTIALS_NAMESPACE),
-            region_name=staticconf.read_string("aws.region"),
-        )
+        # when running clusterman's CLI, we want to use folks' personal AWS creds
+        # so that we don't need to distribute service user creds (and since we can't
+        # use Pod Identity outside of k8s :p)
+        if os.getenv("AWS_PROFILE"):
```
jfongatyelp: I think gating on AWS_PROFILE might bite us when trying to move the actual clusterman instances to pod identity since there's no actual AWS_PROFILE, and i think even `AWS_PROFILE=default` would fail since it's not *technically* a profile to use default session creds. (unless we update this check to accept `AWS_PROFILE=''` as truthy too?)

Could we maybe give ourselves an easy way to transition by also checking USE_POD_IDENTITY envvar exist or something hacky like that?

jfongatyelp (approved):

clusterman/aws/client.py
nemacysts: gah, I had tunnel-visioned on getting the CLI working and totally forgot that we also needed pod identity to work here - lemme refactor this a bit

clusterman/aws/client.py
nemacysts: fixed :)

clusterman/aws/client.py
nemacysts: this should work now!

jfongatyelp (approved):

PR #321, opened 2023-06-01
Upgrading clusterman image to Ubuntu Jammy

wilmer05:
### Description

Upgrading clusterman image to Ubuntu jammy. Updated python version to 3.8 with ppa since the minimum version at Jammy is 3.10

[COMPINFRA-2048](https://jira.yelpcorp.com/browse/COMPINFRA-2048)

### Details

* Ubuntu 22.04 [has not support for python3.6 ](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json). 

* Ubuntu xenial support was removed as most of our clients are already in at least bionic

* tar.gz files were updated for clusterman_signals to use python3.8

* Python version updated to 3.8


package/itest/ubuntu.sh
```diff
@@ -43,18 +43,12 @@ apt-get install -y --force-yes python3.7 python3-pip python3-yaml
 # we also install python3-distutils here to avoid issues on newer ubuntus
 # where disutils isn't included with python (and even though clusterman depends on it, the right
 # version isn't installed in this itest container)
-if  [ "${DISTRIB_CODENAME}" != "xenial" ]; then
 apt-get install -y --force-yes python3.7-distutils
```
gmdfalk: we upgrade to python3.8 elsewhere but not here?

wilmer05: right, thanks for the catch!

wilmer05: updated

clusterman/migration/event.py
```diff
@@ -71,7 +72,7 @@ def _load_instance_type_target(target: str) -> str:
     return target
 
 
-CONDITION_TARGET_LOADERS: Dict[ConditionTrait, Callable[[str], Union[str, int, List[str]]]] = {
+CONDITION_TARGET_LOADERS: Dict[ConditionTrait, Callable[[str], Union[Any, str, int, List[str]]]] = {
```
nemacysts: is this right? `Union` of `Any` and any other type is more or less the same as just having `Any`

requirements-minimal.txt
```diff
@@ -7,10 +7,10 @@ colorama
 humanfriendly
 jsonpickle
 kubernetes
-matplotlib>=3.1.1
+matplotlib==3.4.2
 mypy-extensions
-numpy>=1.16
-packaging
+numpy>=1.20.3
```
nemacysts, line 12: imo, it's fine to remove this lower-bound - we're not really likely to downgrade dependencies on a whim and if we do ever have to downgrade, we'd likely be setting an upper-bound instead (and adding a comment as to why we can't use newer versions of lib X)

requirements-minimal.txt
```diff
@@ -7,10 +7,10 @@ colorama
 humanfriendly
 jsonpickle
 kubernetes
-matplotlib>=3.1.1
+matplotlib==3.4.2
```
nemacysts: this should be unpinned in this file (and pinned in requirements.txt)

(https://github.com/Yelp/requirements-tools#recommended-requirements-setup-for-applications for a quick intro on what all these requirements-* files are for :))

requirements-minimal.txt
```diff
@@ -7,10 +7,10 @@ colorama
...
 mypy-extensions
-numpy>=1.16
-packaging
+numpy>=1.20.3
+packaging==23.1
```
nemacysts: same as above, this should be unpinned in this file (and pinned in requirements.txt)

requirements-minimal.txt
wilmer05: thanks for the tip, I didn't know about this doc :D

requirements-minimal.txt
wilmer05: left it flexible again :)

Makefile
```diff
@@ -187,3 +185,4 @@ debug:
 .PHONY:
 upgrade-requirements:
 	upgrade-requirements --python python3.7
+	upgrade-requirements --python python3.8
```
nemacysts, line 187: we should probably update the python version in the preceding line instead :) (we really only expect services/apps to run with a single python version)

debian/rules
```diff
@@ -8,10 +8,10 @@ endif
 # Use the dpkg version comparison algorithm (which is easily in reach) to compare the build
 # system version to the version from which python3-distutils is required (18.04 and higher),
 # and conditionally populate a substitution variable which is referenced in the control file.
-# On bionic, where Ubuntu's python3.7 is used, this is a virtual package for python3-distutils
-# (in bionic-updates). On jammy, this is a real package providing distutils for python3.7 from deadsnakes.
+# On bionic, where Ubuntu's python3.8 is used, this is a virtual package for python3-distutils
+# (in bionic-updates). On jammy, this is a real package providing distutils for python3.8 from deadsnakes.
 ifeq ($(shell (. /etc/os-release && dpkg --compare-versions $$VERSION_ID "ge" "18.04" && echo yes || echo no)),yes)
-    extra_substvars = -VbionicOrLater:Depends="python3.7-distutils"
+    extra_substvars = -VbionicOrLater:Depends="python3.8-distutils"
```
nemacysts: i think we might be able to drop these shenanigans now and unconditionally include distutils in the control file, no?

nemacysts (commented):
🔥

debian/rules
wilmer05: yes, you're right, thanks for the catch!

Makefile
wilmer05: right, thanks!

clusterman/migration/event.py
wilmer05: right, fixed it now

ilkinmammadzada (approved):

PR #324, opened 2023-06-14
updated kubernetes client lib to 24.2.0

ajayOO8:
### Description

clusterman is using kubernetes=10.0.1 which is way older,  bumping this lib to. 24.2.0 to support new API ref

### DIFF
--- a/clusterman/kubernetes/kubernetes_cluster_connector.py
+++ b/clusterman/kubernetes/kubernetes_cluster_connector.py
@@ -24,8 +24,8 @@ import arrow
 import colorlog
 import kubernetes
 import staticconf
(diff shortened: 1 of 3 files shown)


piax93 (approved):

gmdfalk:
LGTM but note that we can only use this with 1.22 clusters. So we can only roll this out in upgraded clusters.

From the [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/): 
```
Note: policy/v1 Eviction is available in v1.22+. Use policy/v1beta1 with prior releases.
```

gmdfalk (approved):

ajayOO8:
closing this in favor of #334

PR #327, opened 2023-06-27
add and emit pool owner metadata for alerting

gmdfalk:
We'd like to have `pool_owner` metadata on each pool for both informational and alerting purposes.
The new attribute on the pool config, `pool_owner`, defaults to `compute_infra`.

Signed-off-by: Max Falk <gfalk@yelp.com>

### Description

Please fill out!

### Testing Done

working correctly.


acceptance/srv-configs/clusterman-clusters/local-dev/default.mesos
```diff
@@ -29,3 +29,4 @@ autoscale_signal:
       minute_range: 10
 
 alert_on_max_capacity: false
+pool_owner: compute_infra
```
nemacysts: nit: needs a trailing newline

clusterman/autoscaler/autoscaler.py
```diff
@@ -181,7 +181,7 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             self.target_capacity_gauge.set(new_target_capacity, {"dry_run": dry_run})
             self.max_capacity_gauge.set(
                 self.pool_manager.max_capacity,
-                {"dry_run": dry_run, "alert_on_max_capacity": self.pool_manager.alert_on_max_capacity},
+                {"dry_run": dry_run, "alert_on_max_capacity": self.pool_manager.alert_on_max_capacity, "team": self.pool_manager.pool_owner},
```
nemacysts: should we `s/team/pool_owner` to keep things consistent between config and metric labeling?

nemacysts (approved):
minor nits, but nothing blocking :)

jfongatyelp (approved):

clusterman/autoscaler/autoscaler.py
gmdfalk: I knowingly set it to `team` because that's what our alertmanager automatically resolves for ticketing/alerting.
If we don't set it to `team` here, we'll have to relabel in the prometheus query for the `max_pool_capacity` alerting.

Still want to keep it at `pool_owner`? @nemacysts @jfongatyelp

nemacysts: oh, that's fine then - might be good to add that in a comment here tho since i'm not sure how many people are aware of that (or maybe i'm just the only one that didn't know this :p)

nemacysts (approved):

clusterman/autoscaler/autoscaler.py
```diff
@@ -181,7 +181,11 @@ def run(self, dry_run: bool = False, timestamp: Optional[arrow.Arrow] = None) ->
             self.target_capacity_gauge.set(new_target_capacity, {"dry_run": dry_run})
             self.max_capacity_gauge.set(
                 self.pool_manager.max_capacity,
-                {"dry_run": dry_run, "alert_on_max_capacity": self.pool_manager.alert_on_max_capacity},
+                {
+                    "dry_run": dry_run,
+                    "alert_on_max_capacity": self.pool_manager.alert_on_max_capacity,
+                    "team": self.pool_manager.pool_owner,
```
nemacysts: @gmdfalk actually, I was just talking to @EmanElsaban about another alert we wanted to use internally and we realized that having a `team` label on all of the clusterman metrics would be useful - thoughts on adding the label to everything else in this PR?

gmdfalk: @nemacysts Yeah, probably makes sense!

In that case, i think we should be able to add the team label to `autoscaler._emit_requested_resource_metrics` and `drainer._emit_draining_metrics` to tag onto most if not all metrics emitted?

nemacysts: ++ - if we have a centralized place that sounds even better!

I'm a little torn about whether or not this would be of use for the drainer metrics: is there any case where we're not at fault for the drainer not working and we'd want pool owners to deal with drainer pages?

nemacysts: although, i suppose it would be useful for filtering regardless :)

PR #338, opened 2023-08-04
Add support to use in-cluster service account

Buffer0x7cd:
### Description
Currently running clusterman inside a pod requires out of band setup since clusterman needs kubeconfig to access k8s api server. We can fix this by making clusterman the already created Service account which is attached to clusterman pod. 
In cases where clusterman is started outside of a pod ( Like clusterman cli), It falls back to using provided kubeconfig file.

### Testing Done
Added tests to confirm that right load function is called while running inside Pod vs running on host.

### DIFF
--- a/clusterman/kubernetes/util.py
+++ b/clusterman/kubernetes/util.py
@@ -63,10 +63,10 @@ class KubeApiClientWrapper:
             """
             https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
             Every pod in k8s gets some default environment variable injected, including KUBERNETES_SERVICE_HOST
-            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
-            a pod vs when it's started on host. For clusterman instances running inside a k8s cluster, we prioritise using
-            K8s Service account since that let us avoid creating any kubeconfig in advance.
-            For clusterman CLI invocation we continue using provided KUBECONFIG file
+            which points to default kuberbetes service. We are using this variable to distinguise between
+            when cluterman is started in a pod vs when it's started on host. For clusterman instances running inside
+            a k8s cluster, we prioritise using K8s Service account since that let us avoid creating any kubeconfig
+            in advance. For clusterman CLI invocation we continue using provided KUBECONFIG file
             """
             if os.getenv("KUBERNETES_SERVICE_HOST"):
                 kubernetes.config.load_incluster_config()
--- a/clusterman/kubernetes/util.py
+++ b/clusterman/kubernetes/util.py
@@ -62,7 +62,7 @@ class KubeApiClientWrapper:
         try:
             """
             https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
-            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
+            Every pod in k8s gets some default environment variable injected, including KUBERNETES_SERVICE_HOST
             which points to K8s service. We are using this variable to distinguise between when cluterman is started in
             a pod vs when it's started on host. For clusterman instances running inside a k8s cluster, we prioritise using
             K8s Service account since that let us avoid creating any kubeconfig in advance.
--- a/clusterman/kubernetes/util.py
+++ b/clusterman/kubernetes/util.py
@@ -64,7 +64,7 @@ class KubeApiClientWrapper:
             https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
             Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
             which points to K8s service. We are using this variable to distinguise between when cluterman is started in
-            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
+            a pod vs when it's started on host. For clusterman instances running inside a k8s cluster, we prioritise using
             K8s Service account since that let us avoid creating any kubeconfig in advance.
             For clusterman CLI invokation we continue using provided KUBECONFIG file
             """
--- a/clusterman/kubernetes/util.py
+++ b/clusterman/kubernetes/util.py
@@ -60,7 +60,18 @@ class KubeApiClientWrapper:
         :param Type client_class: k8s client class to initialize
         """
         try:
-            kubernetes.config.load_kube_config(kubeconfig_path, context=os.getenv("KUBECONTEXT"))
+            """
+            https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
+            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
+            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
+            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
+            K8s Service account since that let us avoid creating any kubeconfig in advance.
+            For clusterman CLI invokation we continue using provided KUBECONFIG file
+            """
+            if os.getenv("KUBERNETES_SERVICE_HOST"):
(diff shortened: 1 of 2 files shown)


clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
...
+            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
+            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
+            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
+            K8s Service account since that let us avoid creating any kubeconfig in advance.
+            For clusterman CLI invokation we continue using provided KUBECONFIG file
+            """
+            if os.getenv("KUBERNETES_SERVICE_HOST"):
+                kubernetes.config.load_incluster_config()
+            else:
+                kubernetes.config.load_kube_config(kubeconfig_path, context=os.getenv("KUBECONTEXT"))
```
gmdfalk, line 74: do we really set `KUBECONTEXT`? i thought the convention is to use `KUBECONFIG`. And imho that should be checked first to prefer overrides, otherwise assume we run in a k8s environment and can use the incluster config.

gmdfalk (approved):

clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
...
-            kubernetes.config.load_kube_config(kubeconfig_path, context=os.getenv("KUBECONTEXT"))
+            """
+            https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
+            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
+            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
+            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
+            K8s Service account since that let us avoid creating any kubeconfig in advance.
+            For clusterman CLI invokation we continue using provided KUBECONFIG file
+            """
+            if os.getenv("KUBERNETES_SERVICE_HOST"):
```
nemacysts, line 71: I think what we (well, other Yelpers) have been doing is checking if `/var/run/secrets/kubernetes.io/serviceaccount` exists and taking that to mean that we're running inside k8s

that said, not a blocker if you prefer this method instead

clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
         :param Type client_class: k8s client class to initialize
         """
         try:
-            kubernetes.config.load_kube_config(kubeconfig_path, context=os.getenv("KUBECONTEXT"))
+            """
+            https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
+            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
+            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
+            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
```
nemacysts: nit:
```suggestion
            a pod vs when it's started on host. For clusterman instances running inside a k8s cluster, we prioritise using
```

clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
         :param Type client_class: k8s client class to initialize
         """
         try:
-            kubernetes.config.load_kube_config(kubeconfig_path, context=os.getenv("KUBECONTEXT"))
+            """
+            https://kubernetes.io/docs/concepts/containers/container-environment/#container-environment
+            Every pod in k8s get some default enviornment variable injected, Including KUBERNETES_SERVICE_HOST
```
nemacysts: nit: 
```suggestion
            Every pod in k8s gets some default environment variable injected, including KUBERNETES_SERVICE_HOST
```

clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
...
+            which points to K8s service. We are using this variable to distinguise between when cluterman is started in
```
nemacysts: it points to the host that the pod is running in, no? not to a k8s Service?

clusterman/kubernetes/util.py
nemacysts: the kubecontext stuff is for the CLI (since we generally point that at a kubeconfig with multiple clusters)

clusterman/kubernetes/util.py
```diff
@@ -60,7 +60,18 @@ def __init__(self, kubeconfig_path: str, client_class: Type) -> None:
...
+            a pod vs when it's started on host. For clusterman services running inside a k8s cluster, We priiotise using
+            K8s Service account since that let us avoid creating any kubeconfig in advance.
+            For clusterman CLI invokation we continue using provided KUBECONFIG file
```
nemacysts: nit:
```suggestion
            For clusterman CLI invocation we continue using provided KUBECONFIG file
```

nemacysts (approved):
lgtm - i have some comments/suggestions, but none are blocking :)

clusterman/kubernetes/util.py
Buffer0x7cd: I think it was added as a way to support `Kubeconfig` with multiple clusters. In clusterman this result into a empty string.
The reason to prefer using `KUBERNETES_SERVICE_HOST` is that, it's guaranteed to be injected by k8s inside a pod. So whenever this env is set we can assume that the clusterman inside a Pod.

clusterman/kubernetes/util.py
Buffer0x7cd: The `load_incluster_config` method does check `/var/run/secrets/kubernetes.io/serviceaccount `and raises `ConfigException` if it doesn't find the service account file, But i went with this since it's guaranteed to injected by k8s and found it simpler than trying `load_incluster_config` first and than trying `load_kube_config` in case of an exception.

clusterman/kubernetes/util.py
Buffer0x7cd: It points to default k8s service, which is used to connect to API server.

clusterman/kubernetes/util.py
nemacysts: oh, I meant that internally we tend to do this check like 
```python
if Path('/var/run/secrets/kubernetes.io/serviceaccount').exists():
    load_incluster_config()
else:
    do_something_else()
```

that said: we should verify that `paasta local-run` doesn't set `KUBERNETES_SERVICE_HOST` since at least in the Mesos days we "lied" about several env vars to make sure that a local-run would behave as similar as possible to running in an actual cluster - otherwise, local-running clusterman won't work as we'd try to use non-existent service account creds :)

clusterman/kubernetes/util.py
nemacysts: ah gotcha, might be worth rewording this a bit to "points to the default `kubernetes` service" since it's a bit ambiguous right now

nemacysts (approved):

clusterman/kubernetes/util.py
Buffer0x7cd: https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/local_run.py#L659
seems like we only overwrite variables for MESOS.

nemacysts: nice!

PR #341, opened 2023-08-17
U/msingh/clusterman 812 cache kubernetes client creation

Buffer0x7cd:
Currently, with each call to reload_state , we create a new instance of k8s client object. This causes constant increase in memory.
This PR make k8s client part of class attribute which let us reuse the existing k8s client without need to create a new client every single time.
More details at: https://yelp.slack.com/archives/CA53K7S68/p1692279792012659


gmdfalk (approved):

clusterman/kubernetes/util.py
```diff
@@ -54,33 +54,36 @@
 
 
 class KubeApiClientWrapper:
+    _client = None
```
nemacysts, line 59: might be worth a comment here re: why this is a classvar now (just to make sure that we can figure out why we did this even if someone moves this line around or something and the git blame no longer points directly at this commit) :)

nemacysts (approved):

PR #348, opened 2023-10-18
Revert previously undeployed changes

nemacysts:
One of the two reverted PRs appears to be causing a memory leak - I have a hunch that it's the k8s client upgrade, but I'm reverting both of these changes just in case until we can figure out what's going on


nemacysts:
actually, it's definitely the k8s upgrade: we never merged the fix from https://github.com/Yelp/clusterman/pull/341

nemacysts:
but that's failing tests, so I'll revert for now anyway

jfongatyelp (approved):

